PR #6, opened 2022-02-03
Solution

Krimat:


Oleksandr563487953 (approved):

tests/test_split_integer.py
```diff
@@ -1,21 +1,52 @@
 from app.split_integer import split_integer
+from random import randint
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
```
IvanRamyk, line 5: This test is very difficult to understand
It's better to just write assertions with concrete values
In this case, it might be:
```
assert sum(split_integer(13, 5)) == 13
```
Do not write tests to assertions that check the same thing

tests/test_split_integer.py
```diff
@@ -1,21 +1,52 @@
...
+    for tested_number in tested_values:
+
+        # testing each possible splitting
+        for pieces in range(1, tested_number + 1):
+
+            splitted_number = split_integer(tested_number, pieces)
+            assert sum(splitted_number) == tested_number
 
 
 def test_should_split_a_number_into_equal_parts_when_value_is_divisible_by_number_of_parts():
```
IvanRamyk, line 18: The same for all tests: just write assertions with concrete values that cover the corresponding test case.

IvanRamyk (changes requested):

MasterpieceElbow (approved):

PR #7, opened 2022-02-03
'Solution'

denyslipin:


nikon4uk (approved):

tests/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    assert sum(split_integer(10, 3)) == 10
 
 
 def test_should_split_a_number_into_equal_parts_when_value_is_divisible_by_number_of_parts():
-    pass
+    assert split_integer(8, 2)[0] == split_integer(8, 2)[1]
```
IvanRamyk, line 9: It's better to check the result implicitly:
```
assert split_integer(8, 2) == [4, 4]
```
This test is more readable

IvanRamyk (changes requested):

tests/test_split_integer.py
denyslipin: I agree.

MasterpieceElbow (approved):

PR #8, opened 2022-02-03
'Solution'

alexandra-halytska:


tests/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    goals = split_integer(10, 3)
```
nikon4uk, line 5: Variable 'goals' is redundant, function can be called just in assertion statement. For example, sum(split_integer(10, 3)) == 10

nikon4uk (changes requested):

Abnormaltype (approved):

IvanRamyk (approved):
Well done!

PR #13, opened 2022-02-03
Solution

IgorOlifirenko:


tests/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    assert sum(split_integer(32, 6)) == 32
 
 
 def test_should_split_a_number_into_equal_parts_when_value_is_divisible_by_number_of_parts():
-    pass
+    assert len(split_integer(25, 5)) == 5
```
TPolina, line 9: I think, this testcase should check if all parts are equal, and not if there amount is equal to number_of_parts. Don't use len, just check if split_integer(25, 5) gives a correct array.

TPolina (changes requested):

tests/test_split_integer.py
IgorOlifirenko: thx) I fixed it.

TPolina (approved):
Well done!

IvanRamyk (approved):
Well done!

PR #14, opened 2022-02-03
Solution

nikon4uk:


Oleksandr563487953 (changes requested):
Variable 'actual' is redundant, function can be called just in assertion statement. For example, sum(split_integer(6, 2)) == 6

alexandra-halytska (approved):

IvanRamyk (approved):

PR #15, opened 2022-02-03
'Solution'

Oleksandr563487953:


serhii56465 (changes requested):
You can do all these tests without this redundancy. Like this https://prnt.sc/26o0s87

Oleksandr563487953:
> You can do all these tests without this redundancy. Like this https://prnt.sc/26o0s87

Yes, i agree, changes provided. But in some cases i leaved 'goals' variable for better readability. For instance, when i need it to access by index (like in 'for' loop).

serhii56465 (approved):

tests/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
...
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    assert sum(split_integer(6, 2)) == 6
 
 
 def test_should_split_a_number_into_equal_parts_when_value_is_divisible_by_number_of_parts():
-    pass
+    goals = split_integer(20, 4)
+
+    for item in goals[:-1]:
```
IvanRamyk, line 11: It's more difficult to read than the following check:
```
assert split_integer == [5, 5, 5, 5]
```
If you can write the result implicitly do it) 
It's easier to understand the test in this case

tests/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
...
+        assert item == 20 / 4
 
 
 def test_should_return_a_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    goals = split_integer(18, 4)
+
+    assert goals[0] == 4
```
IvanRamyk, line 18: This test should cover the case when you divide an integer not by four, but by one part

tests/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
...
-    pass
+    goals = split_integer(37, 9)
+    goals_to_compare = goals[:]
+
+    assert goals_to_compare == sorted(goals)
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts():
-    pass
+    goals = split_integer(5, 7)
```
IvanRamyk, line 29: Also here is better to compare the result with the concrete list
```
split_integer(5, 7) == [0, 0, 1, 1, 1, 1, 1]
```

tests/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
...
 def test_should_return_a_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    goals = split_integer(18, 4)
+
+    assert goals[0] == 4
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal():
-    pass
+    goals = split_integer(37, 9)
```
IvanRamyk, line 22: It's better to compare the result with the concrete list

IvanRamyk (changes requested):

tests/test_split_integer.py
Oleksandr563487953: Sorry, i can not understand. Do i need to divide 18 by 1, like split_integer(18, 1) ?

tests/test_split_integer.py
Oleksandr563487953: Changes provided, in future i will try to make tests more simple.

tests/test_split_integer.py
IvanRamyk: Right 👍

IvanRamyk (changes requested):

tests/test_split_integer.py
Oleksandr563487953: Changes provided. Thanks for explanation, the test function name was not so clear for me.

IvanRamyk (approved):
Well done!

PR #17, opened 2022-02-03
Solution

anitaposliedova:


Oleksandr563487953 (approved):

tests/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_a_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    assert split_integer(6, 1) == [6]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal():
-    pass
+    assert split_integer(10, 3) == [2, 4, 4]
```
IvanRamyk, line 17: This test is failed with correct reference because it doesn't  satisfy the following condition:
> - The difference between the max and min number in the array should be <= 1

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #18, opened 2022-02-03
"Solution"

Geksus:


alix-mak (approved):

tests/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    assert sum(split_integer(32, 6)) == 32
 
 
 def test_should_split_a_number_into_equal_parts_when_value_is_divisible_by_number_of_parts():
-    pass
+    assert split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
```
IvanRamyk, line 9: Value is not divisible by the number of the parts here.

IvanRamyk (changes requested):

tests/test_split_integer.py
Geksus: But why? 5 + 5 +5 +5 +6 +6 = 32

IvanRamyk: This means division without a remainder, for example:
30 : 6 == 5
```
split_integer(30, 6) == [5, 5, 5, 5, 5, 5]
```

IvanRamyk (changes requested):

tests/test_split_integer.py
Geksus: I finally understood what you ment. Fixed

MasterpieceElbow (approved):

PR #24, opened 2022-05-30
Provide autotests

IvanRamyk:


.github/workflows/test.yml
```diff
@@ -1,9 +1,6 @@
 name: Test
```
danylott, line 1: You forgot to change README.md tests/ -> app/
image

tests/test_tests.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_could_split_on_different_parts(monkeypatch):
+    def split_on_equal_part(value: int, number_of_parts: int):
+        return [value // number_of_parts] * number_of_parts
+
+    monkeypatch.setattr(split_integer, "split_integer", split_on_equal_part)
+
+    test_result = pytest.main(["../app/test_split_integer.py"])
+    print(test_result)
```
danylott, line 13: remove this please

tests/test_tests.py
```diff
@@ -0,0 +1,54 @@
...
+    assert test_result.value == 1
```
danylott, line 14: I'd add here message. Because it is not obvious:
```suggestion
    assert test_result.value == 1, "This solution `split_on_equal_part` should not pass all tests"
```

danylott (approved):
Please fix suggestions

PR #35, opened 2022-06-01
Solution

NOSSUDAK:


.github/workflows/test.yml
```diff
@@ -9,6 +9,9 @@ jobs:
     steps:
```
KatiukhaO, line 9: folder .github/workflows, dont necessary to commit, it`s needless foldef, 
create commit and pul request without this folder

your tests (solution) are ok!)

KatiukhaO (approved):

PR #36, opened 2022-06-01
'Solution'

DmytroSvirsa:


Y-Havryliv (approved):

app/test_split_integer.py
```diff
@@ -2,20 +2,45 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    value = 20
+    number_of_parts = 4
+
+    result = split_integer(value, number_of_parts)
+
+    assert result.count(value / number_of_parts) == 4
```
TPolina, line 19: this test doesn't check that the number is split into equal parts. The result [5, 4, 6, 5] will pass your test because there are 4 parts

TPolina (changes requested):

MasterpieceElbow (approved):

PR #41, opened 2022-06-01
'Solution'

Aleksei-Isaev:


Y-Havryliv (approved):

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    assert split_integer(10, 2) == [5, 5]
```
TPolina, line 5: this test is the same as the second one. Instead, here you should check that sum of parts equals the initial value

TPolina (changes requested):

MasterpieceElbow (approved):

PR #51, opened 2022-06-02
All done

Flashmobber:


app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
...
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    assert sum(split_integer(17, 4)) == 17
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    result = split_integer(6, 2)
+    assert abs(min(result) - max(result)) <= 1
```
NOSSUDAK, line 10: You can also do it in this way: assert len(set(split_integer(36, 6))) == 1

NOSSUDAK (approved):

app/test_split_integer.py
TPolina: you should check if the number is split into equal parts, but you do something different. Use variant that Ilias proposed or even better check the result explicitly: `split_integer(6, 2) == [3, 3]`

TPolina (changes requested):

MasterpieceElbow (approved):

PR #55, opened 2022-06-02
Solution

Chardona:


Y-Havryliv (approved):

app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
...
 
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    assert split_integer(value=99, number_of_parts=1) == [99]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal():
-    pass
+    assert split_integer(value=32, number_of_parts=8) == \
+           sorted(split_integer(value=32, number_of_parts=8))
```
TPolina, line 18: you have to test whether unequal parts are sorted, but 32 is equally divided by 8

TPolina (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    assert sum(split_integer(value=32, number_of_parts=6)) == 32
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    assert split_integer(value=16, number_of_parts=4) == [4, 4, 4, 4]
```
MasterpieceElbow, line 9: Add test when the value is not equally divided by number_of_parts

MasterpieceElbow (changes requested):

app/test_split_integer.py
Chardona: I need to change this test? Add new assert? Write new test?

MasterpieceElbow (approved):

PR #66, opened 2022-06-05
Solution

Vasyl-Smutok:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    assert split_integer(32, 1) == [32]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal():
-    pass
+    assert split_integer(32, 6) == sorted(split_integer(32, 6))
```
IvanRamyk, line 17:
```suggestion
    assert split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
```
It's better to specify the expected value

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #86, opened 2022-07-26
MY_Solution

shevchenko-anastasiia:


Abnormaltype (changes requested):
all tests should are pass

app/split_integer.py
```diff
@@ -1,7 +0,0 @@
-def split_integer(value: int, number_of_parts: int) -> list:
```
IvanRamyk, line 1: Keep the function in the repository. Otherwise, tests won't pass

IvanRamyk (changes requested):

PR #89, opened 2022-08-02
Solutions

InnaKushnir:


app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
...
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    parts = split_integer(7, 1)
+    assert parts[0] == 7
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal():
-    pass
+    parts = split_integer(17, 4)
+    assert parts == sorted(parts)
```
IvanRamyk, line 20:
```suggestion
    assert split_integer(17, 4) == [4, 4, 4, 5]
```
Explicit is better than implicit due to zen of python 
https://peps.python.org/pep-0020/

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #90, opened 2022-08-04
Solution

AlenOl:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
...
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    actual = split_integer(17, 4)
+    assert sum(actual) == 17
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    actual = split_integer(16, 4)
+    assert actual == [4, 4, 4, 4]
```
IvanRamyk, line 11:
```suggestion
    assert split_integer(16, 4) == [4, 4, 4, 4]
```

IvanRamyk: It's better to use an explicit comparison

IvanRamyk (changes requested):

app/test_split_integer.py
AlenOl: Or should I change these conditions in all tests? In the training videos, Danya says that this is the best practice.

app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    actual = split_integer(17, 4)
+    assert sum(actual) == 17
```
Abnormaltype, line 6:
```suggestion
    assert split_integer(17, 4) == 17
```

Abnormaltype (changes requested):
please, fix all cases

Abnormaltype (approved):
Well done!

Abnormaltype (changes requested):
all tests should are pass

Abnormaltype (approved):
Well done!

PR #91, opened 2022-08-09
solution

hikehikehike:


app/test_split_integer.py
```diff
@@ -2,20 +2,40 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    value, number_of_parts = 8, 1
```
Abnormaltype, line 5: it's unnecessary here, please remove it, also fix it in other cases

app/test_split_integer.py
```diff
@@ -2,20 +2,40 @@
...
+
+    result = split_integer(value, number_of_parts)
```
Abnormaltype, line 7: the same here, in this case, it's necessary to use variable

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #100, opened 2022-08-23
Solution

volodymyr-vereshchak:


Abnormaltype (changes requested):
if you want to split for a different variable, please use `pytest.parametrize`

Abnormaltype (approved):
Well done!

PR #104, opened 2022-08-26
Complete task

Glasis9:


app/test_split_integer.py
```diff
@@ -2,20 +2,46 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    result = split_integer(8, 1)
```
IvanRamyk, line 5: this variable is extra. Just put function call to the assertion

app/test_split_integer.py
```diff
@@ -2,20 +2,46 @@
...
+    assert sum(result) == 17
+
+    result = split_integer(32, 6)
+    assert sum(result) == 32
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    result = split_integer(6, 2)
+    assert len(result) == 2
```
IvanRamyk, line 20: You check the number of parts instead of their values

IvanRamyk (changes requested):

app/test_split_integer.py
Glasis9: corrected

IvanRamyk (approved):
Good job!

PR #105, opened 2022-08-29
Solution

VladyslavKlevanskyi:


app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    value = 32
```
IvanRamyk, line 5: These variables are extra
```
```
You can use just one expression which is easier to read

IvanRamyk (changes requested):

app/test_split_integer.py
VladyslavKlevanskyi: Ok. I was just trying to write it in detail, as the tutorial video said.

IvanRamyk (approved):
Well done!

PR #106, opened 2022-08-29
'Solution'

Kev1nXD:


app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
...
+    value = 6
+    number_of_parts = 2
+    assert len(split_integer(value, number_of_parts)) == number_of_parts
 
 
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    value = 8
+    number_of_parts = 1
+    assert split_integer(value, number_of_parts) == [8]
```
IvanRamyk, line 19: Do not use extra variables in tests
```suggestion
    assert split_integer(8, 1) == [8]
```

IvanRamyk (changes requested):

PR #107, opened 2022-08-29
Solution

Kev1nXD:
I deleted old repository and pushed this new commit because I constantly was getting 403 error after each push with old one

### DIFF
--- a/app/test_split_integer.py
+++ b/app/test_split_integer.py
@@ -2,15 +2,11 @@ from app.split_integer import split_integer
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    value = 6
-    number_of_parts = 2
-    assert sum(split_integer(value, number_of_parts)) == value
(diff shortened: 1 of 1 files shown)


app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
...
+    value = 6
+    number_of_parts = 2
+    assert sum(split_integer(value, number_of_parts)) == value
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    value = 6
+    number_of_parts = 2
+    assert len(split_integer(value, number_of_parts)) == number_of_parts
```
IvanRamyk, line 13: Do not use extra variables. They make code less readable

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #113, opened 2022-09-17
'Solution'

YuliyaGerasymniuk:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    assert split_integer(17, 1) == [17]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal():
-    pass
+    assert split_integer(3, 1) == sorted(split_integer(3, 1))
```
IvanRamyk, line 17: Use explicit expected values to make code easier to read 
```suggestion
    assert split_integer(3, 1) == [3]
```
Also, this test case doesn't actually check if the list is sorted (it consists of only one element)

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #117, opened 2022-09-19
Solution

lksndr307:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
...
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    goals = split_integer(6, 2)
+    assert goals == [3, 3]
 
 
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    goals = split_integer(17, 4)
+    assert goals == [4, 4, 4, 5]
```
IvanRamyk, line 16:
```suggestion
    assert split_integer(17, 4) == [4, 4, 4, 5]
```
do not create extra variables. in this case, it makes code less readable

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #118, opened 2022-09-21
checklist and pass test

DmytroSvirsa:


checklist.md
```diff
@@ -0,0 +1,25 @@
...
+def get_full_name(first_name: str, last_name: str) -> str:
+    return f"{first_name} {last_name}"
+```
+
+Bad example:
+```python
+def get_full_name(x: str, y: str) -> str:
+    return f"{x} {y}"
+```
+
```
Abnormaltype, line 19: unnecessary here, add point about don't create unnecessary variables
ex:
```
def test_():
    x = func()
    assert x == [10]
```
 
 would be better
 
```
 def test_():
    assert func() == [10]
```

Abnormaltype (changes requested):

Abnormaltype (approved):

PR #126, opened 2022-09-25
'Solution'

TetyanaPavlyuk:


DmytroSvirsa (changes requested):
You have such a great coments(mostly), just use them, you dont need so mach code for so simple tests

DmytroSvirsa (approved):

PR #128, opened 2022-09-26
Solution

AVyha:


app/test_split_integer.py
```diff
@@ -1,21 +1,63 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    assert len(split_integer(8, 4)) == 4
 
 
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    assert len(split_integer(8, 8)) == 8
```
mltdll, line 51: And this one should check if, for example, `split_integer(8, 1) == [8]`.

mltdll (changes requested):
I think you should re-read the task and implement the simplest possible solution. No need to even introduce any variables.

mltdll (changes requested):
Oops, forgot to reject.
UPD: Or did I?

AVyha:
I know that i dont need to do many tests but i tried to use interesting thing that i saw in videos. I think i dont need to delete this tests) For me they looks cool.
P.S. Incorrect tests fixed.

mltdll (approved):

Nattalli (approved):
Good job!

PR #132, opened 2022-09-27
Develop

hai-le-di:


app/test_split_integer.py
```diff
@@ -2,20 +2,23 @@
...
-    pass
+    assert sum(split_integer(17, 4)) == 17
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    min_num = min(split_integer(17, 4))
+    max_num = max(split_integer(17, 4))
+    difference = max_num - min_num
+    assert difference <= 1
```
DmytroSvirsa, line 12: 17 is not divisible by 4, we want to test if we split `number` in a number of parts that is a divisor of `number` - will we gut the list with equal parts?

app/split_integer.py
```diff
@@ -1,5 +1,11 @@
 def split_integer(value: int, number_of_parts: int) -> list:
     parts = []
+    if number_of_parts > value:
+        for num in range(value):
+            parts.append(1)
+        for num in range(number_of_parts - value):
+            parts.append(0)
+        return parts
```
DmytroSvirsa, line 8: you don't need to change it

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #134, opened 2022-09-28
Solution

kateryna3k:


AriohBelskij (approved):
n1

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    assert sum(split_integer(8, 1)) == 8
```
mltdll, line 5: I'd argue that it would be a more useful test to split number into more parts than one in this case.

mltdll (changes requested):

mltdll (approved):

mltdll (approved):

MasterpieceElbow (approved):
GJ!

PR #135, opened 2022-09-28
Solution

taras-andruschenko:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_or_the_parts_should_be_equal_to_value():
-    pass
+    assert split_integer(8, 1) == [8]
```
mltdll, line 5: Uhm, this is not the check for this test. Did you read the name carefully?

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    assert split_integer(6, 2) == [3, 3]
 
 
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    assert split_integer(17, 4) == [4, 4, 4, 5]
```
mltdll, line 13: Same here.

mltdll (changes requested):

app/test_split_integer.py
mltdll: Actually, now that I have read the name carefully, I can see that there's a typo. I'll report that...

taras-andruschenko: totally agree

mltdll (approved):

MasterpieceElbow (approved):
GJ!

PR #144, opened 2022-10-02
Develop

rostyak:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value():
-    pass
+    assert sum(split_integer(13, 3)) == 13
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    assert [4, 4, 4, 4] == split_integer(16, 4)
```
mltdll, line 9: I'd suggest to turn those checks around so that expected result is after the tested value, because it seems to be an unspoken universal agreement on format.
In other tests as well.

mltdll (changes requested):

mltdll (approved):

Y-Havryliv (approved):
GJ!

PR #147, opened 2022-10-03
Complete!

ketstap162:


app/split_integer.py
```diff
@@ -5,3 +5,6 @@ def split_integer(value: int, number_of_parts: int) -> list:
         parts.append(value // parts_left)
         value -= next_number
     return parts
+
+
+print(split_integer(5, 10))
```
mltdll, line 10: You forgot to remove this line.

app/test_split_integer.py
```diff
@@ -2,20 +2,33 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value():
-    pass
+    assert sum(split_integer(32, 6)) == 32
```
mltdll, line 5: There are some problems with grouping tests like that:
- Assertions chained inside one function won't execute if the previous one fails. You generally should use `pytest.mark.parametrize()` if you're using pytest.
- Adding more assertions for the same functionality is fine, but not strictly necessary. If the cases you test for are not executed in the function in different ways, you testing for just one of them is often considered enough.

mltdll (changes requested):

mltdll (approved):

Y-Havryliv (approved):
Well done!

PR #149, opened 2022-10-04
Solution

linaSam:


app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
...
+    result = split_integer(10, 3)
+
+    assert result == [3, 3, 4]
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts():
-    pass
+    result = split_integer(5, 6)
+
+    assert result == [0, 1, 1, 1, 1, 1]
```
mltdll, line 31: You don't need to introduce a variable for result. It's easier to read if you don't.

mltdll (changes requested):

app/test_split_integer.py
linaSam: Thanks!

mltdll (approved):

Y-Havryliv (approved):
GJ!

PR #154, opened 2022-10-04
Solution

Yaroslav-Stetsko:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    assert split_integer(6, 2) == [3, 3]
 
 
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    assert split_integer(8, 4) == [2, 2, 2, 2]
```
HStakhiv, line 13: I think that this test should have 1 as the second parameter

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value():
-    pass
+    assert sum(split_integer(8, 1)) == 8
```
HStakhiv, line 5: I think 1 as the second parameter in this test is a bad idea. What do you think?

HStakhiv (changes requested):

HStakhiv (approved):
:D

app/test_split_integer.py
```diff
@@ -1,21 +1,72 @@
...
+            id="sum of the parts should be equal to value"
+        )
+    ]
+)
+def test_correct_values_for_sum(value, number_of_parts, return_result):
+    assert sum(split_integer(value, number_of_parts)) == return_result
 
-def test_should_add_zeros_when_value_is_less_than_number_of_parts():
-    pass
+# def test_sum_of_the_parts_should_be_equal_to_value():
```
Y-Havryliv, line 55: delete comments

Y-Havryliv (changes requested):

app/test_split_integer.py
Yaroslav-Stetsko: Done

Y-Havryliv (approved):
GJ!

PR #161, opened 2022-10-04
Solution

FranchukS:


app/test_split_integer.py
```diff
@@ -2,20 +2,23 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value():
-    pass
+    assert 17 == sum(split_integer(17, 4))
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    result = split_integer(40, 5)
```
Vasyl-Poremchuk, line 9: I think you do not need to create a variable for this test (check the `checklist` file for this task). You can write:
`assert split_integer(40, 5) == [8, 8, 8, 8, 8]`

app/test_split_integer.py
```diff
@@ -2,20 +2,23 @@
...
 
 
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    assert 1 == len(split_integer(25, 1))
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal():
-    pass
+    result = split_integer(32, 6)
```
Vasyl-Poremchuk, line 18: Same here

app/test_split_integer.py
```diff
@@ -2,20 +2,23 @@
...
+    assert result == sorted(result)
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts():
-    pass
+    result = split_integer(4, 9)
```
Vasyl-Poremchuk, line 23: And here

Vasyl-Poremchuk (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_is_divisible_by_parts():
-    pass
+    assert [8, 8, 8, 8, 8] == split_integer(40, 5)
 
 
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    assert 1 == len(split_integer(25, 1))
```
TPolina, line 13: when you write tests you need to make them as obvious as possible. You can write: `split_integer(25, 1) == [25]` and that's all. No need to do additional actions (`len`, `count`, and so on). There are no sense is this

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal():
-    pass
+    assert [5, 5, 5, 5, 6, 6] == sorted(split_integer(32, 6))
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts():
-    pass
+    assert 5 == split_integer(4, 9).count(0)
```
TPolina, line 21: same with `count` as with `len`. Write simple tests

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_part_equals_to_a_value_when_slitting_into_one_part():
-    pass
+    assert 1 == len(split_integer(25, 1))
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal():
-    pass
+    assert [5, 5, 5, 5, 6, 6] == sorted(split_integer(32, 6))
```
TPolina, line 17: this test is incorrect. You need to test that the result is sorted by function `split_integer` but in your test, you receive the result of this function and then sort it so of course it will be sorted and you don't check anything

TPolina (changes requested):

TPolina (approved):
Well done!

PR #173, opened 2022-10-05
'Solution'

Nikyxa:


app/test_split_integer.py
```diff
@@ -2,20 +2,32 @@
...
-def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
 
+def test_should_split_into_equal_parts_when_value_is_divisible_by_parts() \
+        -> None:
+    parts = split_integer(16, 4)
 
-def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert parts == [4, 4, 4, 4]
```
VitaliyBychevoy, line 14: in all tests can be somethink like: 
assert split_integer(16, 4) == [4, 4, 4, 4], "some error message"

TPolina (approved):
Well done! But try to avoid `\` in your code. I approved the solution because these long test names are ours and that's why you have problems. We will fix it

PR #178, opened 2022-10-05
Solution

pavlo-kapitanets:


YevheniiDavydenko (approved):
nice!
but you have conflicts with master branch and idk its mistake or no, so take your approve from me)

app/test_split_integer.py
```diff
@@ -1,21 +1,21 @@
 from app.split_integer import split_integer
 
 
-def test_sum_of_the_parts_should_be_equal_to_value():
-    pass
+def test_sum_of_the_parts_are_equal_to_value() -> None:
+    assert split_integer(17, 4) == [4, 4, 4, 5]
```
TPolina, line 5: in this test you need to get sum or the function result and check that it is equal to the initial value

TPolina (changes requested):

TPolina (approved):
Good job!

PR #180, opened 2022-10-05
Solved the task

Cerne13:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(32, 3)) == 32
```
VitaliyBychevoy, line 5: add error message everywhere

VitaliyBychevoy (changes requested):

app/test_split_integer.py
Cerne13: As far as I understood, we don't use parametrization here (we have separate functions in the boilerplate code).
So we have very informative tracebacks if an assertion error is thrown (with the name of the function, diff etc.).
I doubt we additionally need to do "with pytest.raises()" or "@pytest.mark.xfail" here...

I guess I'll try to pass the solution to mentors as is.

TPolina (approved):
Well done!

app/test_split_integer.py
TPolina: yes, in this task it is not obligatory

PR #187, opened 2022-10-11
'Solution'

IvanGLS:


app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    sum_equal = split_integer(12, 25)
+
+    assert sum(sum_equal) == 12
```
IvanRamyk, line 7: do not create an extra variable
```suggestion
    assert sum(split_integer(12, 25)) == 12
```

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #193, opened 2022-10-19
Solution

Amirammi:


DmytroSvirsa (changes requested):
in this tests, you can do every test with just an assert statement

DmytroSvirsa (approved):

PR #203, opened 2022-11-02
Split

bythewaters:


app/split_integer.py
```diff
@@ -5,3 +5,5 @@ def split_integer(value: int, number_of_parts: int) -> list:
         parts.append(value // parts_left)
         value -= next_number
     return parts
+
+split_integer(10, 6)
```
IvanRamyk, line 9: please, don't leave your test calls where you just writing tests

app/test_split_integer.py
```diff
@@ -1,21 +1,88 @@
...
+                          number_of_parts)
 
+    assert (
+        sum(parts) == value
+    ), f"Sum of the {parts} should be equal to {value}"
 
-def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
 
+@pytest.mark.parametrize(
```
IvanRamyk, line 22: do not use `parametrize` when you want to test only one set of parameters

IvanRamyk: just write regular tests

IvanRamyk (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,29 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    parts = split_integer(32, 6)
+
+    assert sum(parts) == 32
```
IvanRamyk, line 7:
```suggestion
    assert sum(parts) == split_integer(32, 6)
```
avoid using extra variables 

fix everywhere

IvanRamyk (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
...
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+
+    assert split_integer(6, 2) == [3, 3]
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+
+    assert split_integer(6, 1)[0] == 6
```
IvanRamyk, line 16:
```suggestion
    assert split_integer(6, 1) == [6]
```
it's better to check the result explicitly

IvanRamyk: fix everywhere

IvanRamyk (changes requested):

DmytroSvirsa (approved):

PR #205, opened 2022-11-04
Solution

SpecializedBaby:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(4, 4) == [1, 1, 1, 1]
```
DmytroSvirsa, line 5: the name of test function is not the sama that the test check

DmytroSvirsa (changes requested):

PR #213, opened 2022-11-24
Solution

Greenmanxd:


app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
...
+    assert (
+        split_integer(32, 6) == sorted(split_integer(32, 6))
+    ), "Result have to be sorted!!!"
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert (
+        split_integer(1, 2) == [0, 1]
+    ), "If value is less than number of parts you have to add zeros!!!"
```
YuriiKindrat, line 31: All tests should be passed

YuriiKindrat (changes requested):

app/test_split_integer.py
Greenmanxd: i dont understand what is the problem

Serhii-Leonenko (approved):
Good job!

PR #214, opened 2022-11-26
Solution

ochernyi:


app/split_integer.py
```diff
@@ -1,7 +1,10 @@
...
     for parts_left in range(number_of_parts, 0, -1):
-        next_number = value // parts_left
-        parts.append(value // parts_left)
-        value -= next_number
+        if value < number_of_parts:
+            parts.append(0)
+        else:
+            next_number = value // parts_left
+            parts.append(value // parts_left)
+            value -= next_number
```
Serhii-Leonenko, line 9: You shouldn't change this code!

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(8, 1)) == 8
```
Serhii-Leonenko, line 5: Use other data for the test, where the count of parts more than one

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(17, 4) == [4, 4, 4, 5]
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(2, 3) == [0, 0, 0]
```
Serhii-Leonenko, line 21: This test isn't correct, the result of the function split_integer must be another because you changed the code in split_integer.py.

Serhii-Leonenko (changes requested):

IvanRamyk (approved):

PR #217, opened 2022-11-29
Solution

Vladyslav7125:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(7, 1) == [7]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(33, 5) == sorted(split_integer(33, 5))
```
IvanRamyk, line 17:
```suggestion
    assert split_integer(33, 5) == [6, 6, 7, 7, 7]
```
> Explicit is Better Than Implicit

IvanRamyk (changes requested):

app/test_split_integer.py
Vladyslav7125: fixed

Vladyslav7125: What about my task?

Vladyslav7125: What about my task?

IvanRamyk (approved):

PR #233, opened 2022-12-22
Solution

sorrow03:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(14, 3)) == 14
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(15, 3)[0] == split_integer(15, 3)[-1]
```
katevasilenko, line 9: I would prefer to see all three result in this case

katevasilenko (changes requested):
Looks good just change a bit one test

katevasilenko (approved):
Thanks for make changes!

PR #249, opened 2023-01-09
Solution

yuliia-stopkyna:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(27, 7)) == 27
```
IvanRamyk, line 5:
```suggestion
    assert split_integer(27, 7) == [3, 4, 4, 4, 4, 4, 4]
```
explicit is better than implicit

IvanRamyk (changes requested):

olena-hul (approved):

PR #259, opened 2023-01-17
Solution

YuriiVataschuk:


ArsenPidhoretskyi (changes requested):
You do not cover the function with the following criteria
1. The difference between the max and min number in the array should be <= 1
2. The array should be sorted ascending (from lowest to highest)

Also, I recommend you use @pytest.mark.parametrize for different test cases

ArsenPidhoretskyi (approved):

ArsenPidhoretskyi:
Looks okay

PR #262, opened 2023-01-25
'Solution'

timyr4iq:


app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 32
+    number_of_parts = 6
```
Vasyl-Smutok, line 6: I think you should not create these variables, you can write it all in one line

Example:
```
```

Please fix it here in other places

Vasyl-Smutok (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #272, opened 2023-02-08
solution

VladyslavOladko:


app/test_split_integer.py
```diff
@@ -2,20 +2,36 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 17
+    number_of_parts = 4
+    result_list_of_function = split_integer(value, number_of_parts)
+    assert (sum(result_list_of_function) == value)
```
Vasyl-Smutok, line 8: don't make it difficult)

```suggestion
    assert (sum(result_list_of_function) ==17)
```

please fix it here and other places

Vasyl-Smutok (changes requested):

PR #278, opened 2023-02-12
Solution

Zoriana-Dvulit:


README.md
```diff
@@ -14,7 +14,8 @@ integer elements:
 You don't need to validate arguments (they are always valid).
 
 Examples:
-```
+``
```
olena-hul, line 17: Do not change readme

app/test_split_integer.py
```diff
@@ -2,20 +2,37 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 6
+    number_of_parts = 2
```
olena-hul, line 6: use numbers directly in assert, no need for extra variables

fix in all places

olena-hul (changes requested):

README.md
Zoriana-Dvulit: Very strange, I didn't change anything in README. In my PyCharm correct 
image

Bohdan-Salamakha (approved):

PR #283, opened 2023-02-15
Dev roman senyshyn py split integer

AndereLion:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
```
MaybeTI, line 5: here you have to check the list sum

MaybeTI (changes requested):

app/test_split_integer.py
AndereLion: Fixed

MaybeTI (approved):

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(32, 6)) == sum([5, 5, 5, 5, 6, 6])
```
Serhii-Leonenko, line 5: if you use sum() here just compare it with sum result no need to use sum again and put the list inside it

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(25, 5).count(5) == 5
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(1, 3) == [0, 0, 1]
```
Serhii-Leonenko, line 13: it isn't correct, read the name of the test again

Serhii-Leonenko (changes requested):

app/test_split_integer.py
AndereLion: Fixed

YuriiKindrat (approved):

PR #300, opened 2023-02-20
Solution

Maytreyaya:


amber-marichi (approved):
looks good

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(15, 1)) == 15
```
Serhii-Leonenko, line 5: Looks good, but it's better to use a bigger value for the number of parts in this test, try to not use similar data, the same you use in test_should_return_part_equals_to_value_when_split_into_one_part

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Well done!

PR #306, opened 2023-02-21
Solution

NazarHladaniuk:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(8, 1) == [8]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(17, 4) == sorted(split_integer(17, 4))
```
inrock1, line 17: I think using the same function in the right part isn't correct.
paste list with digit

inrock1 (changes requested):

app/test_split_integer.py
inrock1: there is a check not only sorting

NazarHladaniuk: I don't think so, it performs the same action, it's just written a little differently.
It meets the requirements of the checklist and fits into one line

inrock1 (approved):

YuriiKindrat (approved):

PR #308, opened 2023-02-21
Solution

destr34:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(8, 1) == [8]
```
ada-krav, line 5: you are not checking sum of the parts

ada-krav (changes requested):

ada-krav (approved):

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(8, 1)) == 8
```
YuriiKindrat, line 5: it seems to me, it is better to divide number on more than one part here

YuriiKindrat (changes requested):

wQuelS (approved):
Good

PR #309, opened 2023-02-21
Solution

ada-krav:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(17, 1) == [17]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(47, 11) == [4, 4, 4, 4, 4, 4, 4, 4, 5, 5, 5]
```
ZhAlexR, line 17: You can use "sorted" here instead of hard-coding the values

ZhAlexR (approved):

viktoria-rybenchuk (approved):

app/test_split_integer.py
ada-krav: I don`t get your idea, can you please elaborate?

ZhAlexR: @ada-krav 
The idea was to use `assert split_integer(47, 11) == sorted(split_integer(47,11))` instead `split_integer(47, 11) == [4, 4, 4, 4, 4, 4, 4, 4, 5, 5, 5]`
But after morning lecture I am not sure that is right case and, maybe, just write `[4, 4, 4, 4, 4, 4, 4, 4, 5, 5, 5]` like you did is better case then `sorted(split_integer(47,11))`.
I suppose when a mentor will check he shares his opinion about this)))

YuriiKindrat: @ZhAlexR u can make this in both ways

YuriiKindrat (approved):

PR #317, opened 2023-02-22
Solution

Danil228:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    parts = split_integer(10, 3)
```
NickSavchyn, line 5: not sure if that matters, but you can make every test in 1 line so it would take less space

NickSavchyn (changes requested):

app/test_split_integer.py
Danil228: done

NickSavchyn (approved):
thank you, well done!

YuriiKindrat (approved):

PR #319, opened 2023-02-22
Solution

Anatolii-Poznyak:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(6, 2)) == 6
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(9, 3)[0] == split_integer(9, 3)[1]
```
YuriiKindrat, line 9: what about third part? no need to call mathod twice here, just write a right result

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #322, opened 2023-02-22
Solution

ostapT:


me1nyk (approved):
Good job and nice idea for 2nd and 3d tests

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert len(set(split_integer(16, 4))) == 1
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert len(split_integer(16, 1)) == 1
```
YuriiKindrat, line 13: no need to check len here, u should check a part that equals to value

YuriiKindrat (changes requested):

wQuelS (approved):
Good

PR #323, opened 2023-02-22
Solution

Litvinenko23:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(17, 4) == [4, 4, 4, 5]
```
ostapT, line 5: i think in this test you should check if sum of parts is equal to value. now you just compare the result of the function with the list of parts

ostapT (changes requested):

ostapT (approved):

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == sum([4, 4, 4, 5])
```
YuriiKindrat, line 5: no need to use sum twice here, u can just write a value at the right side

YuriiKindrat (changes requested):

wQuelS (approved):
Good

PR #324, opened 2023-02-22
Solution

VladyslavCherkashyn:


alexlega (approved):

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(17, 4) == [4, 4, 4, 5]
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
```
YuriiKindrat, line 21: are u sure about this test? value should be less than divisor

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #325, opened 2023-02-22
Solution

ghost:


app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(21, 4)) == 21
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
```
dogmoony, line 9: this test should check just case when number splited on equal parts.

app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(23, 1) == [23]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(21, 4) == sorted(split_integer(21, 4))
```
dogmoony, line 18: you could write this by yourself, without using function sorted, as it will take more time

dogmoony (changes requested):

myroslavatrschk (approved):

Serhii-Leonenko (approved):
Good job!

PR #327, opened 2023-02-22
'Solution'

alexlega:


Yulia-Mala (approved):
I am not sure but I guess that descriptive messages after assert are also required. I'll approve but I advise you to add it.

wQuelS (approved):
Good

PR #333, opened 2023-02-22
solution

Yulia-Mala:


app/test_split_integer.py
```diff
@@ -1,21 +1,16 @@
...
-    pass
-
-
-def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+@pytest.mark.parametrize(
+    "number,parts,result",
+    [
+        pytest.param(9, 3, [3, 3, 3], id="equal parts expected if possible"),
+        pytest.param(5, 1, [5], id="same value expected when part = 1"),
```
skmiwa1, line 9: bene!

skmiwa1 (approved):

Serhii-Leonenko (changes requested):
Your solution is ok but do it according to the initial functions, please.

PR #336, opened 2023-02-22
'Solution'

ykopatko:


app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value, number_of_parts = 10, 4
+    result = split_integer(value, number_of_parts)
+    assert sum(result) == value
```
wQuelS, line 7: This can be done in one line. No need to make so many additional variables, since those tests would be pretty much easy to read anyway. Fix everywhere

YuriiKindrat (approved):

PR #372, opened 2023-03-30
Solution

dominyevhen:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(17, 1) == [17]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(17, 4) == sorted(split_integer(17, 4))
```
EdAlekseiev, line 17: You can't use the result of the function you're testing as the expected result. Hardcode expected result as in other methods.

EdAlekseiev (changes requested):

Bohdan-Salamakha (approved):

PR #384, opened 2023-04-14
Solution

DmytroBondariev:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(32, 6)
+    assert sum(result) == 32
```
vasylhnatiuk, line 6: Why do you sum the result? We must not change the function's output during testing.

vasylhnatiuk (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #391, opened 2023-04-26
Solution of Pytests

svyshcho:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
...
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    result = split_integer(32, 6)
+    assert result == [5, 5, 5, 5, 6, 6]
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    result = split_integer(5, 8)
+    assert result == [0, 0, 0, 1, 1, 1, 1, 1]
```
Bohdan-Salamakha, line 26: in this case u don't need to create a separated variable for 1-time use
instead of
```
result = split_integer(6, 2)
assert result == [3, 3]
```
do
```
assert split_integer(6, 2) == [3, 3]
```
etc.

Bohdan-Salamakha (changes requested):
v1

app/test_split_integer.py
svyshcho: Done

Bohdan-Salamakha (approved):

PR #392, opened 2023-04-29
Solution

romkapomka12:


app/test_split_integer.py
```diff
@@ -2,20 +2,23 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    parts = split_integer(8, 1)
+    assert sum(parts) == 8
```
Bohdan-Salamakha, line 6: u don't need to create a separated variable in this case
just do it in one line

app/test_split_integer.py
```diff
@@ -2,20 +2,23 @@
...
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(8, 1) == [8]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    parts = split_integer(17, 4)
+    assert sorted(parts) == [4, 4, 4, 5]
```
Bohdan-Salamakha, line 19: u don't need to create a separated variable in this case
just do it in one line

app/test_split_integer.py
```diff
@@ -2,20 +2,23 @@
...
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    parts = split_integer(2, 6)
+    assert parts == [0, 0, 0, 0, 1, 1]
```
Bohdan-Salamakha, line 24: u don't need to create a separated variable in this case
just do it in one line

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #393, opened 2023-05-02
it gives me file not found error

chipapapo:
so maybe it's again, that my python version isn't right and on server it'll do its job


vasylhnatiuk (approved):
Good job!

PR #410, opened 2023-05-20
Py-split_integer

alexgubarev88:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(11, 3) == [3, 4, 4]
```
ArsenPidhoretskyi, line 5: test_sum
You tested other info

ArsenPidhoretskyi (changes requested):

app/test_split_integer.py
Vasyl-Smutok: not fixed

Vasyl-Smutok, line 5: change the test to check the sum

Vasyl-Smutok (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #411, opened 2023-05-20
solution

IvanKorshunovE:


IvanKorshunovE:
I can't understand whether I did everything correctly. Should I have used @pytest.mark.parametrize?

app/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    parts = split_integer(50, 6)
+
```
Vasyl-Smutok, line 6: I think no need to create variable 

Write it in one line

Vasyl-Smutok (changes requested):

app/test_split_integer.py
IvanKorshunovE: Ok.

Vasyl-Smutok (approved):
Great 👍

PR #432, opened 2023-06-08
Initial commit

thorium-zero:


thorium-zero:
To anyone, who would review this, I suggest to pay attention to my first test, I made an error there, cuz I thought to use template lists as my "expected", but then I forgot my initial intension, took another direction and forgot to change my setup parameters 🤪

PR #436, opened 2023-06-12
Solution

YevheniiIovchu:


app/test_split_integer.py
```diff
@@ -2,20 +2,36 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    value = 16
+    number_of_parts = 4
+    every_integer = value // number_of_parts
+    split = split_integer(value, number_of_parts)
+    for number in split:
+        assert number == every_integer
```
DmytroTsehelnyk, line 17: Why do you need so complicated tests? Try to rewrite your asserts in one line

app/test_split_integer.py
```diff
@@ -2,20 +2,36 @@
...
+    assert split == sorted(split)
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    value = 4
+    number_of_parts = 6
+    difference = number_of_parts - value
+    split = split_integer(value, number_of_parts)
+    assert split == [0] * difference + [1] * value
```
DmytroTsehelnyk, line 37: Look at how asserts work. You should state what the result of the test will be, not write the test itself.

DmytroTsehelnyk (changes requested):

DmytroTsehelnyk (approved):
GJ!

Oleg-s-s (approved):
:+1:

PR #438, opened 2023-06-13
Solution

AntonShpakovych:


Kirontiko (approved):
GJ!

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(6, 2) == [3, 3]
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert len(split_integer(8, 1)) == 1
```
Oleg-s-s, line 13: What if function return wrong value? `[8] != [3]` but `len()`  is

Oleg-s-s (changes requested):

YuriiKindrat (approved):

PR #439, opened 2023-06-13
PyTests (first commit)

IhorVoskoboinikov:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(32, 6)) == 32
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert len(split_integer(6, 2)) == 2
```
ant-komarov, line 9: it's more like you check number of parts, but not their equality

ant-komarov (approved):

app/test_split_integer.py
Oleg-s-s: This part is wrong, that's true

Oleg-s-s (changes requested):

app/test_split_integer.py
Oleg-s-s: Also you forgot to mark checklist in the table as done

ant-komarov (approved):

PR #443, opened 2023-06-13
Solution

Po-SvinSkipodla:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(32, 6)) == 32
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(40, 5).count(8) == 5
```
Tarasidze, line 9: Interesting decision. I like it)

Tarasidze (approved):
GJ!

app/test_split_integer.py
Oleg-s-s, line 9: What if split_integer() unpredictably return [8, 8, 8, 8, 8, 0]

Oleg-s-s (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🐇

PR #446, opened 2023-06-13
Solution

volodymyr-komarnyckyi:


app/test_split_integer.py
```diff
@@ -1,21 +1,34 @@
...
+import unittest
 
-def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+from app.split_integer import split_integer
 
 
-def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+class TestSplitInteger(unittest.TestCase):
```
oleksiikolii, line 6: same, with pytest you don't need to make class

app/test_split_integer.py
```diff
@@ -1,21 +1,34 @@
-from app.split_integer import split_integer
-
+import unittest
```
oleksiikolii, line 1: You need to use PyTest according to Readme *hint: you don't need to import anything

oleksiikolii (changes requested):

app/test_split_integer.py
volodymyr-komarnyckyi: Oh, didn't pay attention

oleksiikolii (approved):
Nice job

Oleg-s-s (approved):
:+1:

PR #453, opened 2023-06-14
Solution

Kasatello:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(10, 1) == [10]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert sorted(split_integer(32, 6)) == [5, 5, 5, 5, 6, 6]
```
OlhaTryhub, line 17: In this line you compare sorted result of function working with obviously sorted list. In this case even result will:

split_integer(32, 6) = [5, 6, 6, 5, 5, 5]

after sorting we won't get AssertionError here in any case. Try to compare unsorted result with sorted result of function working in 17th line.

OlhaTryhub (changes requested):

OlhaTryhub (approved):

PR #455, opened 2023-06-14
Solution

vkrasnovyd:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert len(split_integer(17, 4)) == 4
```
YukihiroSM, line 5: not length, but sum of the parts should be equal to starting number

YukihiroSM (changes requested):

PR #462, opened 2023-06-18
Solution split integer

MarkShapka:


app/test_split_integer.py
```diff
@@ -2,20 +2,27 @@
...
+    value = 10
+    number_of_parts = 2
+    actual = split_integer(value, number_of_parts)
+    assert actual == [5, 5]
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    actual = split_integer(8, 1)
+    assert actual[0] == 8
```
Dimosphen1, line 18: It would be better not to refer to the first element of the list in the assertion clause but rather assert the list as a whole. It will check that there are no excessive elements in the list as well.

Dimosphen1 (approved):
Everything looks fine, one minor comment was left.

PR #466, opened 2023-06-20
solution

vlodek69:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(47, 5) == [9, 9, 9, 10, 10]
```
ArsenPidhoretskyi, line 5: test_sum_of_the_parts
Test name is not the same what tested inside

ArsenPidhoretskyi (changes requested):

vlodek69:
Changed other tests also to better reflect the test names. But for me it seems completely unnecessary, tests just loose functionality (e.g. `assert split_integer(47, 5) == [9, 9, 9, 10, 10]` already checks that 9+9+9+10+10 = 47 AND checks that list is sorted and that number of elements is correct).

ArsenPidhoretskyi (approved):
Well done!

PR #487, opened 2023-07-15
Solution

pnakongit:


app/test_split_integer.py
```diff
@@ -2,20 +2,37 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 32
+    number_of_parts = 6
```
vsmutok, line 6: I think no need to create variable if it use only once

vsmutok (changes requested):

app/test_split_integer.py
pnakongit: Noted! Many thx  !

vsmutok (approved):
Good Job 👍

PR #490, opened 2023-07-17
Implemented test functions

lubart2517:


app/test_split_integer.py
```diff
@@ -2,20 +2,27 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(20, 6)) == 20
+    assert sum(split_integer(8, 2)) == 8,\
+        "Sum of the parts should be equal to start integer value"
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
```
pavlejviki, line 10: You can simplify the test further  by checking that the returned list is as expected.
assert split_integer(4, 2)  == [ 2, 2 ]

app/test_split_integer.py
```diff
@@ -2,20 +2,27 @@
...
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(8, 3) == sorted(split_integer(8, 3)),\
+        "Parts should be sorted when they are not_equal"
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(4, 6)[0] == 0,\
```
pavlejviki, line 27: it would be more effective to check the entire output list instead of checking only the first element

app/test_split_integer.py
```diff
@@ -2,20 +2,27 @@
...
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(4, 2)[0] == split_integer(4, 2)[1]
+    assert split_integer(8, 2)[0] == split_integer(8, 2)[1],\
+        "Parts should be equal when value divisible by parts"
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(8, 1)[0] == 8, \
```
pavlejviki, line 17: it's generally good to verify the entire output of a function in your test

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #506, opened 2023-08-03
Solution

ip-ilya:


app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
...
+    assert split_integer(10, 1) == [10]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    value, number_of_parts = 16, 5
+
+    expected_res = split_integer(value, number_of_parts)
+
+    assert expected_res == sorted(expected_res)
```
pavlejviki, line 21: there is no need to create additional variables. you can write everything in one line

app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
...
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    value, number_of_parts = 3, 5
+
+    expected_res = split_integer(value, number_of_parts)
+
+    assert (len(expected_res) == 5) and (0 in expected_res)
```
pavlejviki, line 29: it would be better to assert the whole list rather than checking its length and the presence of zero separately. 

Avoid creating variables.

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #508, opened 2023-08-03
Solution

nschcur:


vasylhnatiuk (changes requested):
image
change it

app/test_split_integer.py
```diff
@@ -2,20 +2,33 @@
...
+    number_of_parts = 5
+    assert sum(split_integer(value, number_of_parts)) == value
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    value = 25
+    number_of_parts = 5
+    assert (all(x == split_integer(value, number_of_parts)[0]
+                for x in split_integer(value, number_of_parts)))
```
ArsenPidhoretskyi, line 14: x is not good name. Rename to number

ArsenPidhoretskyi, line 14: And also it is not good to call function 6 times
Just call it once and check than len(set(items)) == 1

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #528, opened 2023-08-23
Solution for py-split-integer

Barsh4ec:


app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
...
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(16, 4) == [4, 4, 4, 4]
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    result = split_integer(10, 1)
+    assert result[0] == 10
```
innyshka, line 14: I think, in this test, you can just use "assert split_integer(10, 1) == [10]"

innyshka (approved):
GJ! Juct correct test_should_return_part_equals_to_value_when_split_into_one_part()

PR #530, opened 2023-08-23
Solution

denys-source:


OlegatorLE (approved):
your solution is correct, but it could be simpler by function-based

PR #532, opened 2023-08-24
Solution

OlegatorLE:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    expected_value = split_integer(6, 2)
+    assert expected_value == [3, 3]
```
Barsh4ec, line 6: it will be better to use just `assert split_integer(6, 2) == [3, 3]` without using unnecessary variables and change it in every other test

Barsh4ec (approved):

Barsh4ec (approved):

PR #533, opened 2023-08-24
Solution

9rosLove:


app/test_split_integer.py
```diff
@@ -1,21 +1,48 @@
...
+            number_of_parts: int,
+            result: list[int]
+    ) -> None:
+        assert split_integer(value, number_of_parts) == result
 
 
-def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+def test_sum_of_the_parts_should_be_equal_to_value() -> None:
+    assert sum(split_integer(2, 2)) == 2
```
maxkatkalov, line 48: to be honest, I don`t think that u need this func here because, for example, when u check  125, 5, [25, 25, 25, 25, 25] and test is passed u already know that sum is correct

maxkatkalov (approved):
Good job in general!

PR #535, opened 2023-08-25
tests added for integer.py

maxkatkalov:


ORENYT (approved):
Well done, nice idea to use parametrize in this task.

PR #538, opened 2023-08-25
Solution

ZaichenkoViktoriia:


app/test_split_integer.py
```diff
@@ -1,21 +1,36 @@
...
-    pass
+    assert split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(2, 4) == [0, 0, 1, 1]
+
+
+def test_should_raise_error_for_type_of_value() -> None:
```
OleksandrMoroz24, line 25: why you add some extra tests for validation agruments? we have condition "You don't need to validate arguments (they are always valid)."

app/test_split_integer.py
```diff
@@ -1,21 +1,36 @@
+import pytest
 from app.split_integer import split_integer
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(17, 4) == [4, 4, 4, 5]
```
OleksandrMoroz24, line 7: you need to check that sum of function result equal to value(17 in your case)

OleksandrMoroz24 (changes requested):

PR #545, opened 2023-08-25
'Solution'

WonnaKoo:


Ihor-MA:
I don`t know is it must have to use @pytest.mark.parametrize or not, but looks ok

Ihor-MA (approved):

PR #547, opened 2023-08-26
py-split(1)

ostboiko:


test_main.py
Bohdan-Salamakha, line 1: remove this empty file

Bohdan-Salamakha (changes requested):
simplify every test
for example:
instead of 
```
    result = split_integer(10, 3)
    assert len(result) == 3
    assert max(result) - min(result) <= 1
    assert sum(result) == 10
```
u can write
```
    assert sum(split_integer(10, 3)) == 10
```

app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(10, 3)) == 10
```
olena-hul, line 5: Fix tests

olena-hul (changes requested):

YukihiroSM (approved):
Good job!

PR #553, opened 2023-08-27
Solution

obohatov:


MarianKovalyshyn (approved):
I think it's not obvious at first how you check if the sum of the parts is equal to the value. I would use method 'sum()'.
And also I think you were supposed to do tests in functions that were defined before, but you just deleted them.

PR #554, opened 2023-08-27
Implemented test split integer

mtiunov:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(8, 2)) == 8
```
olena-hul, line 5: Fix tests

olena-hul (changes requested):

PR #557, opened 2023-08-28
solution

ivanzh13:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert len(split_integer(32, 6)) == 6
```
vasylhnatiuk, line 5: you must use here sum() instead of len()

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
but also result of this function must be another

PR #566, opened 2023-09-10
Solution

fotomassimo:


app/test_split_integer.py
```diff
@@ -2,20 +2,23 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (sum(split_integer(17, 4)) == 17)
```
fsocie7y, line 5: remove redundant parentheses in all asserts

app/test_split_integer.py
```diff
@@ -2,20 +2,23 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert (split_integer(5, 1) == [5])
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    actual = split_integer(17, 4)
```
fsocie7y, line 17: u should write other logic in this case
u should check if numbers sorted

fsocie7y (changes requested):

PR #576, opened 2023-09-19
solution

Shum430:


app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 10
+    number_of_parts = 3
+    parts = split_integer(value, number_of_parts)
```
pavlejviki, line 7: no need to create additional variables

app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
...
+    assert sum(parts) == value
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    value = 15
+    number_of_parts = 3
+    parts = split_integer(value, number_of_parts)
+    assert all(part == value // number_of_parts for part in parts)
```
pavlejviki, line 15: It's generally a good practice in testing to compare the actual result with an expected value directly.you can simply compare with the expected list

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #584, opened 2023-09-29
'Solution'

nsdNite:


viktoria-rybenchuk (changes requested):
all tests should passed,fix it

app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
...
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(42, 3)) == 42
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    parts = split_integer(42, 7)
+    assert all(part == parts[0] for part in parts)
```
viktoria-rybenchuk, line 10: you can simplify the test, just write what you expect to receive after execution this function

viktoria-rybenchuk (changes requested):

app/test_split_integer.py
nsdNite: is it a bad practise to make such test? just curious

Bohdan-Salamakha (approved):

PR #585, opened 2023-09-29
Solution

valikdeb:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
...
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(32, 6) == sorted(split_integer(32, 6))
+    assert split_integer(1201, 3) == sorted(split_integer(1201, 3))
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(2, 3) == [0, 1, 1]
+    assert split_integer(1, 4) == [0, 0, 0, 1]
```
Bohdan-Salamakha, line 26: leave only 1 assert in every test
u don't need to test the same logic twice, it's redundant

Bohdan-Salamakha (changes requested):

viktoria-rybenchuk (changes requested):
All tests should pass

PR #589, opened 2023-10-04
Solution

MatsalakO:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(32, 6)
```
pavlejviki, line 5: no need to create additional variable. remove in all tests

pavlejviki (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #606, opened 2023-10-18
Solution

mykolamateichuk:


app/test_split_integer.py
```diff
@@ -2,20 +2,33 @@
...
+    result_list = split_integer(2, 1)
+    assert (
+        len(result_list) == 1 and result_list[0] == 2
+    ), "Should return part equals to value when split into one part!"
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    result_list = split_integer(13, 7)
+    sorted_result_list = split_integer(13, 7)
```
vshvanska, line 25: Are you sure you are actually checking if the list is sorted? In my opinion, there is no need to create separate variables here. Item number one in the checklist

vshvanska (changes requested):

fsocie7y (approved):
well done ^_^

PR #612, opened 2023-10-18
my test

DenPrislipskyi:


app/test_split_integer.py
```diff
@@ -2,20 +2,22 @@
...
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(6, 2) == [3, 3]
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    result = split_integer(17, 4)
+    assert len(result) == 4 and result[0] == 4
```
YuliaHladyshkevych, line 15: This test should check if the value is equal to return part when number_of_parts = 1, for example split_integer(8, 1).

app/test_split_integer.py
```diff
@@ -2,20 +2,22 @@
...
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(13, 5) == sorted(split_integer(13, 5))
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
```
YuliaHladyshkevych, line 23: This test should check when value is less than number_of_parts, for example split_integer(3, 6).

YuliaHladyshkevych (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    total = split_integer(6, 2)
```
fsocie7y, line 5: u can write it in one line

fsocie7y (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,22 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    # total = split_integer(6, 2)
+    # assert sum(total) == 6
```
fsocie7y, line 6: please remove commetns 🙉

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #621, opened 2023-10-19
Solution

RabenkoYevhenii:


fsocie7y (approved):
approved, but i think we can prettier indentation)

PR #623, opened 2023-10-19
Solution

skoval7:


Lyutillis (approved):

app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
...
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(8, 1) == [8]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    result = split_integer(11, 3)
+    assert result == [3, 4, 4]
```
fsocie7y, line 18: u can remove one assert and write it in one line)

app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(8, 1) == [8]
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
```
fsocie7y, line 9: it better to use for ex. 36 here cause in according to task u must had smthn like this `split_integer(36, 6) == [6, 6, 6, 6, 6, 6]`

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #625, opened 2023-10-19
Implement py-split-integer

SevKrok:


fedorov-s-od (approved):
Good job! It's strange but we both wrote the same condition for the last test lol

fsocie7y (approved):
well done ^_^

PR #638, opened 2023-10-27
Solution

okien1:


app/test_split_integer.py
```diff
@@ -2,20 +2,46 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 20
+    number_of_parts = 3
```
pavlejviki, line 6: no need to create additional variables

app/test_split_integer.py
```diff
@@ -2,20 +2,46 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    value = 21
+    number_of_parts = 3
+
+    result = split_integer(value, number_of_parts)
+
+    if value % number_of_parts:
```
pavlejviki, line 17: based on the test's name, you're only testing the scenario where value is divisible by number_of_parts, so the if condition is not needed.

app/test_split_integer.py
```diff
@@ -2,20 +2,46 @@
...
+    value = 20
+    number_of_parts = 3
+
+    assert sum(split_integer(value, number_of_parts)) == value
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    value = 21
+    number_of_parts = 3
```
pavlejviki, line 13: no need to create additional variables

app/test_split_integer.py
```diff
@@ -2,20 +2,46 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    value = 20
+    number_of_parts = 1
+
+    result = split_integer(value, number_of_parts)
+
+    assert result[0] == value
```
pavlejviki, line 29: when writing tests, it's  clearer to compare the expected output directly rather than individual elements of the output.

compare the result with the expected output list

app/test_split_integer.py
```diff
@@ -2,20 +2,46 @@
...
+
+    result = split_integer(value, number_of_parts)
+    sorted_result = sorted(result)
+    assert result == sorted_result
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    value = 5
+    number_of_parts = 8
```
pavlejviki, line 43: no need to create additional variables

app/test_split_integer.py
```diff
@@ -2,20 +2,46 @@
...
+    assert result[0] == value
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    value = 20
+    number_of_parts = 3
+
+    result = split_integer(value, number_of_parts)
+    sorted_result = sorted(result)
```
pavlejviki, line 37: no need to create additional variables

pavlejviki (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,33 @@
...
+    assert result == sorted_result
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+
+    result = split_integer(5, 8)
+
+    assert result == [0, 0, 0, 1, 1, 1, 1, 1]
```
vsmutok, line 34:
```suggestion
    assert split_integer(5, 8) == [0, 0, 0, 1, 1, 1, 1, 1]
```

vsmutok: fix everywhere

vsmutok (changes requested):

vsmutok (approved):
Great 👍

PR #640, opened 2023-10-28
Solution

frezworx:


app/test_split_integer.py
```diff
@@ -2,20 +2,41 @@
...
+        == sorted(split_integer(25, 8))
+    ), "Parts should be sorted when they are not equal"
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    result = split_integer(5, 11)
+    assert (
+        len(result) == 11
+    ), "The number of parts must be correct"
```
vsmutok, line 35:
```suggestion
    assert (len(result) == 11), "The number of parts must be correct"
```

vsmutok: fix indentation, 

you can use `black` to have it do it automatically

vsmutok (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,36 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(10, 3)) == 10, \
+        "Sum of the parts should be equal to value"
```
vsmutok, line 6: better to use `()` then `\`

vsmutok (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,38 @@
...
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (
+        sum(split_integer(10, 3)) == 10
+    ), "Sum of the parts should be equal to value"
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    parts = split_integer(6, 3)
```
viktoria-rybenchuk, line 11: Don't create unnecessary variables for simple tests. Read checklist before pushing your solution

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #649, opened 2023-11-08
'Solution'

Illya-Maznitskiy:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(8, 1) == [8]
```
Dimosphen1, line 5: You should test that the `sum` of parts equals to the initial value. Rewrite this test to obtain a list of several values and then apply `sum` to it, on the right side leave the required integer, not list. Your test body should contain one line with an `assert` statement as well.

Dimosphen1 (changes requested):
Several changes were requested.

Polyakiv-Andrey (approved):
Good job!

Dimosphen1 (approved):
LGTM, well done!

PR #652, opened 2023-11-13
Develop

AndriiHomeniuk1:


app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(10, 3)
+    assert sum(result) == 10, (
```
```suggestion
    assert sum(split_integer(10, 3)) == 10, (
```

I think  no need create variable, if it uses  only once

vsmutok: Please fix everywhere

vsmutok (changes requested):

app/test_split_integer.py
AndriiHomeniuk1: Yes sure, fixed it.
Thank you

ArsenPidhoretskyi (approved):
Well done!

PR #653, opened 2023-11-14
Solution_1

Taran21V:


app/test_split_integer.py
```diff
@@ -2,20 +2,37 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 21
+    number_of_parts = 3
+    assert sum(split_integer(value, number_of_parts)) == value, (
+        "Sum of the parts should be equal to value!"
+    )
```
NikOneZ1, line 9: Why do you need this variables? The same for other tests
```suggestion
    assert sum(split_integer(21, 3)) == value, (
        "Sum of the parts should be equal to value!"
    )
```

EdAlekseiev (approved):
Good job

PR #662, opened 2023-11-19
Solution

weuis:


app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 15
+    number_of_parts = 5
+    result = split_integer(value, number_of_parts)
+    assert sum(result) == value
```
vsmutok, line 8:
```suggestion
    assert sum(split_integer(15, 5)) == 15
```
I think that it is not necessary to create a variable if it is used only once

vsmutok: fix everywhere

vsmutok (changes requested):

vsmutok (approved):
Good Job 👍

PR #664, opened 2023-11-20
Solution

Artem-user-19:


app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
...
+    result = split_integer(value, number_of_parts)
+    assert result == sorted(result)
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    value = 3
+    number_of_parts = 5
+    result = split_integer(value, number_of_parts)
+    assert result == [0, 0, 1, 1, 1]
```
Dimosphen1, line 36: Rewrite tests so that there would be only an `assert` statement inside of each test without allocation of additional variables to be more concise.

Dimosphen1 (changes requested):
Several changes were requested.

Dimosphen1 (changes requested):
Several changes were requested.

viktoria-rybenchuk (approved):
GJ!🚀

PR #669, opened 2023-11-23
Solution

Marinel444:


app/test_split_integer.py
```diff
@@ -2,20 +2,26 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(13, 5)
```
viktoria-rybenchuk, line 5: Don't create unnecessary variables for simple tests. Read checklist

viktoria-rybenchuk (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,22 @@
...
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(13, 5)) == 13
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    result = split_integer(8, 2)
+    assert result[0] == result[1]
```
Dimosphen1, line 10: You could consider an approach of not creating additional variables and testing against a particular output, e.g.:
```suggestion
    assert split_integer(8, 2) == [4, 4]
```
But your approach is valid as well.

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration.

PR #672, opened 2023-11-26
Split-Integer-pytest-without-class

MykolaMazh:


app/test_split_integer.py
```diff
@@ -2,20 +2,23 @@
...
-    pass
+    assert split_integer(844, 93) == sorted(split_integer(844, 93))
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(25, 32) == [
+        0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1,
+        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+        1, 1, 1, 1, 1, 1, 1, 1]
```
Dimosphen1, line 24: As a general recommendation, there is no need to use large numbers in test cases when it is not a necessary part of the test case itself. It would be appropriate to shorten the compared list and the magnitude of numbers overall for this task.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #705, opened 2024-01-07
Solution

eLQeR:


app/test_split_integer.py
```diff
@@ -1,21 +1,81 @@
...
+@pytest.mark.parametrize(
+    "value,number_of_parts,result",
+    [
+        (12, 1, [12])
+    ]
+)
+def test_should_return_part_equals_to_value_when_split_into_one_part(
+        value: int,
+        number_of_parts: int,
+        result: int
```
YukihiroSM, line 47: Result is not int in this case (you are providing [12] this is the list

YukihiroSM: This should be fixed everywhere.

YukihiroSM (changes requested):

vasylhnatiuk (approved):

app/test_split_integer.py
```diff
@@ -1,21 +1,81 @@
...
 from app.split_integer import split_integer
 
 
-def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+@pytest.mark.parametrize(
+    "value,number_of_parts",
+    [
+        (9, 3)
+    ]
```
NEkropo1, line 10: in case of 1 value it looks strange)

NEkropo1 (approved):
LGTM!

PR #706, opened 2024-01-09
Develop

Zonda001:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(8, 1)
+    assert result == [8]
```
vsmutok, line 6: tests should match the name of test,  for example this test should check the sum

vsmutok: Also no need  to create `result` variable as it is only used once

vsmutok (changes requested):

PR #709, opened 2024-01-10
Solution

ArturPoltser:


aLEKS-e3 (approved):
good!!

NEkropo1 (changes requested):
It's actually good, you only need to remove slashes on end, since it's not pythonic.
But I suggest you to try parametrize in these cases, with messages thrown as ids.

PR #712, opened 2024-01-13
Solution

aLEKS-e3:


app/test_split_integer.py
```diff
@@ -3,7 +3,7 @@
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
     assert sum(split_integer(10, 4)) == 10, \
-        f"Expected sum of the parts "\
+        "Expected sum of the parts "\
         f"to be equal 10, but got {sum(split_integer(10, 4))}."
```
NEkropo1, line 7: remove this weird symbol at the end of all lines(`\`)

NEkropo1 (changes requested):

app/test_split_integer.py
aLEKS-e3: pycharm automatically puts them. but is it better with parentheses?

NEkropo1: yup it is

NEkropo1 (approved):
LGTM!

PR #713, opened 2024-01-14
py-split-integer

Vladyslav-Pliashechnyk:


app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17,\
+        f"Sum of parts = {sum(split_integer(17, 4))}, "\
+        f"but it should be equal to 17"
```
NEkropo1, line 7: You don't need f-strings in cases, where you don't use links to objects, also remove this endslash.

NEkropo1 (changes requested):

NEkropo1 (changes requested):
I meant you should remove f-strings and use literal strings, where you didn't mention objects xD
And, even though we don't actually need messages in these cases, since test names are good, better point would be to leave them, so we can faster understand where the problem is.

app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    assert sum(split_integer(17, 4)) == 17
+    assert sum(split_integer(17, 4)) == 17, \
+        f"Sum of parts = {sum(split_integer(17, 4))}, \
+        but it should be equal to 17"
```
NEkropo1, line 7:
```suggestion
    assert sum(split_integer(17, 4)) == 17, (
        f"Sum of parts = {sum(split_integer(17, 4))},
        but it should be equal to 17"
        )
```
something like this )

NEkropo1 (changes requested):

PR #727, opened 2024-01-20
resolved

DonSosiska:


tests/test_tests.py
NazarSheremet, line 1: You cannot change the files of auto tests

NazarSheremet (changes requested):
Read this [doc](https://github.com/mate-academy/py-split-integer/blob/master/checklist.md) and change your test according to best practice

PR #729, opened 2024-01-20
solution: split integer

sacrosanctry:


app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
...
+        f"Sum of parts should be equal to 36 "
+        f"but is {sum(split_integer(36, 7))}"
+    )
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(42, 6) == [7] * 6, (
+        f"Number should be divided into equal parts if it is divisible "
+        f"by number of parts but actual is {(split_integer(36, 7))}"
```
vsmutok, line 14: why `split_integer(36, 7)` ??

vsmutok (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(36, 7)) == 36, (
+        f"Sum of parts should be equal to 36 "
+        f"but is {sum(split_integer(36, 7))}"
+    )
```
vsmutok, line 8:
```suggestion
    result = sum(split_integer(36, 7))
    assert result == 36, (
        f"Sum of parts should be equal to 36 "
        f"but is {result}"
    )
```

it's better to create a variable so you don't call the same function twice

vsmutok (changes requested):

PR #747, opened 2024-01-28
Develop

MaxOperchuk:


app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(20, 5)
+    assert sum(result) == 20
```
mdubyna, line 6: u can write it in one line

mdubyna (approved):

app/test_split_integer.py
JohnnyWalker010: I confirm)

JohnnyWalker010 (approved):

PR #753, opened 2024-01-28
'Solution'

artemlitvenko:


app/test_split_integer.py
```diff
@@ -2,20 +2,36 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 10
+    parts = 3
+    result = split_integer(value, parts)
+    assert sum(result) == value, "sum(result) == value"
```
tskopenko, line 8: by the cheklist its not correct
git_check_1

tskopenko, line 8: For what u return `"sum(result) == value"`? In error it will be visible for user.
In other func as well. Not only here.

tskopenko (changes requested):
Fix this pls.

PR #761, opened 2024-01-29
Solution

Bulleter06:


artemgrishko (changes requested):
Rewrite tests with @pytest.mark.parametrize

PR #764, opened 2024-01-30
Develop

Serhii-Hrynyshyn:


EdAlekseiev (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
+            ), "Sum of the parts should be equal to 32"
```
YukihiroSM, line 6: You are not really checking, that the sum is equal to 32, you are checking the parts. into which number is being splitted. 
What about using: 

```suggestion
    assert (sum(split_integer(32, 6)) == 32
            ), "Sum of the parts should be equal to 32"
```

YukihiroSM (changes requested):
Else looks nice to me!

EdAlekseiev (approved):
Good job

PR #768, opened 2024-02-04
Develop

SerheiZhus:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(8, 1)
+    assert sum(result) == 8
```
```suggestion
    assert sum(split_integer(8, 1)) == 8
```

vsmutok: no need create variable if it use only once

vsmutok (changes requested):

PR #779, opened 2024-02-07
Devops

volry:


app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
...
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    parts = split_integer(17, 4)
+    assert parts == sorted(parts)
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(1, 1) == [1]
```
NazarSheremet, line 25: This test is incorrect, need to test if zeros are added when the value is less than a number

NazarSheremet (changes requested):
Need to change one test

YuriiSmolii (approved):
LGTM

PR #784, opened 2024-02-12
solution

yaaszp:


app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 17
+    number_of_parts = 4
+    result = split_integer(value, number_of_parts)
+    assert sum(result) == value
```
NazarSheremet, line 8: No need to create variables for `value, number_of_parts, result`. 
```suggestion
    assert sum(split_integer(17, 4)) == 17
```

NazarSheremet (changes requested):
Rewrite tests folowing the example, which I left in comment

app/test_split_integer.py
NazarSheremet: I understand, it was done this way in the video for a better explanation

NazarSheremet (approved):
Good job!

PR #786, opened 2024-02-13
Solution

MykhailoIvchenko:


EdAlekseiev (changes requested):
Fix tests

MykhailoIvchenko:
One test failes because the function is not correct. This one case test_should_add_zeros_when_value_is_less_than_number_of_parts
Or I understand it wrong? There shouldn't be all zeros?

EdAlekseiev (approved):
Good job

PR #794, opened 2024-02-16
'Solution'

oboliziuk:


app/test_split_integer.py
```diff
@@ -2,20 +2,26 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(6, 2)
+    assert sum(result) == 6
```
```suggestion
    assert sum(split_integer(6, 2)) == 6
```
no need to create `result` variable if it use only once

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #807, opened 2024-02-27
Solution

dkibalenko:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(23, 5) == [4, 4, 5, 5, 5]
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(8, 11)[0:3] == [0, 0, 0]
```
Dimosphen1, line 21: Write tests that are not overly complicated and avoid using slicing, compare to the list that is returned directly

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert len(set(split_integer(32, 8))) == 1
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert 32 in split_integer(32, 1)
```
Dimosphen1, line 13: Compare to the list that is returned directly, avoid eventual checks since `split_integer(32, 1)` could return `[32, 1]` and this check would be successful

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert len(set(split_integer(32, 8))) == 1
```
Dimosphen1, line 9: Compare to the list that is returned directly without `list(set(` operation

Dimosphen1 (changes requested):

PR #812, opened 2024-03-04
Solution

Roman-Sokolov-V:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert len(split_integer(5, 3)) == 3
```
Dimosphen1, line 5: This test should test `sum`, not `len`. It should equal 5 in this case

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #816, opened 2024-03-06
'Solution'

KravchenkoSerhiii:


app/test_split_integer.py
```diff
@@ -2,20 +2,40 @@
...
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    results = split_integer(3, 5)
+    assert sorted(results) == results
+
+    results = split_integer(5, 7)
+    assert sorted(results) == results
+
+    results = split_integer(10, 15)
+    assert sorted(results) == results
```
Dimosphen1, line 41: It would be enough to add one assert in each test. Besides that, do not allocate additional variables such as `result`, compare function result in the assert directly. Each test should be implemented in one line

Dimosphen1 (changes requested):

PR #826, opened 2024-03-12
Solution

ILyakhova:


app/test_split_integer.py
```diff
@@ -2,20 +2,36 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 10
+    number_of_parts = 3
+    parts = split_integer(value, number_of_parts)
+    assert sum(parts) == value
```
NazarSheremet, line 8: Based on  clean code recommendations, will be better to do:
```suggestion
    assert split_integer(10, 3) == [3, 3, 4]
```
image

NazarSheremet (changes requested):
Need to implement suggestions for all tests

YuriiSmolii (approved):
LGTM

PR #840, opened 2024-03-18
Develop

SHCHERBANV:


app/test_split_integer.py
```diff
@@ -1,21 +1,32 @@
-from app.split_integer import split_integer
+from .split_integer import split_integer
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value, number_of_parts = 17, 4
+    parts = split_integer(value, number_of_parts)
+    assert sum(parts) == value
```
NazarSheremet, line 7: Based on clean code recommendations, will be better to do:

```suggestion
    assert split_integer(17, 4) == [4, 4, 4, 5]
```
image

NazarSheremet (changes requested):
Need to implement suggestions for all tests. Also no need to add file with tests results

PR #854, opened 2024-03-23
Solution

Vova-Stoliar:


app/test_split_integer.py
```diff
@@ -2,20 +2,47 @@
...
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    value = 6
+    number_of_parts = 2
+
+    split_integer_output = split_integer(value=value,
+                                         number_of_parts=number_of_parts)
+
+    assert split_integer_output == [3, 3]
```
LiudmylaKulinchenko, line 22: I'd suggest you to avoid creating additional variables and to try to rewrite in one line
In these cases you don't reuse variables and logic is pretty simple so in this case no need to create them

LiudmylaKulinchenko (changes requested):

PR #861, opened 2024-03-26
Solution for task split integer

oleksandra-shershen:


Boyok1337 (approved):
GJ!

app/test_split_integer.py
```diff
@@ -2,20 +2,32 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(8, 3)
```
YuriiKindrat, line 5: no need to create variable here

app/test_split_integer.py
```diff
@@ -2,20 +2,32 @@
...
+    assert sum(result) == 8
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    result = split_integer(12, 4)
+    assert len(result) == 4
+    assert max(result) - min(result) <= 1
+    assert sorted(result) == result
```
YuriiKindrat, line 13: it seems to me, that test for dividing in to equal parts should look like this 
`assert max(result) - min(result) == 0`
also it will be much eadier to do like this 
`assert split_integer(12, 4) == [3,3,3,3]`
this test will also cover all needed cases

app/test_split_integer.py
```diff
@@ -2,20 +2,32 @@
...
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    result = split_integer(3, 5)
+    assert len(result) == 5
+    assert max(result) - min(result) <= 1
+    assert sorted(result) == result
+    assert result.count(0) == 5 - 3
```
YuriiKindrat, line 33: same is here

YuriiKindrat (changes requested):

fsocie7y (approved):
GJ 💯

PR #864, opened 2024-03-27
Solution

Line181:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (sum(split_integer(50, 5)) == 50
+            ), "sum of your elements should equal 50"
```
Dimosphen1, line 6: It would be enough to leave only an assert statement to not spread it onto multiple lines. The test name is self-descriptive enough. Please, apply similar changes to all functions
```suggestion
    assert sum(split_integer(50, 5)) == 50
```

Dimosphen1 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (sum(split_integer(50, 5)) == 50)
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
```
pavlopro, line 9: It would be enough to leave only an assert statement to not spread it onto multiple lines. The test name is self-descriptive enough. 

Please, apply similar changes to all functions

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #876, opened 2024-04-01
split integer

mikhailyemets:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(32, 6)) == 32
```
Oleksl888, line 5: add error messages here and below

Oleksl888 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,26 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    parts = split_integer(32, 6)
+    assert sum(parts) == 32, \
```
Oleksl888, line 6: Use round braces to break lines instead of `\`

Oleksl888 (changes requested):

app/test_split_integer.py
mikhailyemets: assert works incorrectly when i use round braces with it..

Oleksl888: From [https://peps.python.org/pep-0008/](Pep8)

> The preferred way of wrapping long lines is by using Python’s implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

Though `\` is acceptable by some, we don't use it here

Oleksl888: > assert works incorrectly when i use round braces with it..
does this work for you?

```assert split_integer(8, 1) == [8], (
        "The value was not returned as a single part when split into one part."
    )

mikhailyemets: In this way it works.
I made a mistake and put into round braces all statement after assert and it provides to errors.

Oleksl888 (approved):
Well done!

PR #877, opened 2024-04-01
Success?

Olevsky-Ivan:


app/test_split_integer.py
```diff
@@ -1,21 +1,54 @@
...
+        ]
+    )
+    def test_sum_of_the_parts_should_be_equal_to_value(
+            self,
+            first_int: int,
+            second_int: int,
+            result: list
+    ) -> None:
+        assert split_integer(first_int, second_int) == \
+               result, f"Sum of the parts should be equal to value {first_int}"
```
vsmutok, line 23: better use `()` then `\`

app/test_split_integer.py
```diff
@@ -1,21 +1,54 @@
...
+
+    def test_should_split_into_equal_parts_when_value_divisible_by_parts(
+            self
+    ) -> None:
+        first_int = 6
+        second_int = 2
+        assert split_integer(first_int, second_int) == [3, 3],\
+            "Test should split into equal parts when value is divisible by"
```
vsmutok, line 31:
```suggestion
        assert split_integer(6, 2) == [3, 3],\
            "Test should split into equal parts when value is divisible by"
```

vsmutok: no need to create variable, if it used only once

vsmutok (changes requested):

EdAlekseiev (changes requested):
If you use pytest.mark.parametrize you should join your test functions into one

EdAlekseiev (approved):

PR #884, opened 2024-04-02
'Solution'

dkotkod:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(8, 1)) == 8
```
Oleksl888, line 5: add error messages here and below

Oleksl888 (changes requested):

Oleksl888 (approved):
Good one!

PR #885, opened 2024-04-03
solution

alonasorochynska:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(6, 2) == [3, 3]
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(17, 4) == [4, 4, 4, 5]
```
Oleksl888, line 13: how does this splits value in one part?

Oleksl888 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(8, 1)) == 8
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
-
-
-def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
```
Oleksl888, line 12: why did you remove this test completely? it must be implemented

Oleksl888 (changes requested):

app/test_split_integer.py
Oleksl888: splits in one part means that value split by 1 will result in a list with len of 1 containing element of value

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #886, opened 2024-04-03
Implement Split integer task

bwaah1:


app/test_split_integer.py
```diff
@@ -2,20 +2,33 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    initial_value = 32
```
Oleksl888, line 5: No need to declare variables if you are going to use them only once

app/test_split_integer.py
```diff
@@ -2,20 +2,33 @@
...
+    number_of_parts = 6
+    assert (sum(split_integer(initial_value, number_of_parts))
+            == initial_value)
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    initial_value = 16
```
Oleksl888, line 12: if you reuse same parameters  in different test they can be placed into a fixture

Oleksl888 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,26 @@
...
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert len(set(split_integer(16, 4))) == 1, \
+        "When value divisible by parts split into equal parts"
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(8, 1)[0] == 8, \
```
Oleksl888, line 15: This test is logically incorrect. What if your function returns [8,8,8] - wrong result but still be valid according to your test

app/test_split_integer.py
```diff
@@ -2,20 +2,26 @@
...
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert (split_integer(16, 4) == sorted(split_integer(16, 4))), \
+        "Parts should be sorted when they are not equal"
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    diff = 10 - 6
+    assert split_integer(6, 10)[:diff] == [0] * diff, \
```
Oleksl888, line 26: Can you check it in a more simple way? Would be easier to compare certain input data and correct return data

Oleksl888 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (sum(split_integer(32, 6)) == 32), \
```
Oleksl888, line 5: Why you put sum expression in parenthesis? Also we don't use `\` for breaking lines as per our code standards - here you actually must use parenthesis

Oleksl888 (changes requested):
Approved

Oleksl888 (changes requested):
changes required

app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
...
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(8, 1) == [8], (
+        "Return part equals to value when split into one part"
+    )
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(16, 4) == sorted(split_integer(16, 4)), (
```
Oleksl888, line 23: the program splits 16 into 4 equal parts - [4,4,4,4]. So the result is always sorted, you need to select different dataset to test this

Oleksl888 (changes requested):

Oleksl888 (approved):
Cool

PR #890, opened 2024-04-03
'Solution'

kssshevchenko:


app/test_split_integer.py
```diff
@@ -1,21 +1,67 @@
...
 
 
-def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+@pytest.mark.parametrize("value, number_of_parts, expected_result", [
+    (8, 1, [8]),
+    (6, 2, [3, 3]),
+    (17, 4, [4, 4, 4, 5]),
+    (32, 6, [5, 5, 5, 5, 6, 6])
+])
```
Oleksl888, line 10: Nice that you going over and beyond using parametrize here, but it is also a good practice to add id/ids for parametrized tests

app/test_split_integer.py
```diff
@@ -1,21 +1,67 @@
 from app.split_integer import split_integer
+import pytest
```
Oleksl888, line 2: what about order of imports?

app/test_split_integer.py
```diff
@@ -1,21 +1,67 @@
...
+                                                   ) -> None:
+    assert value == sum(expected_result)
 
 
-def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+@pytest.mark.parametrize("value, number_of_parts, expected_result", [
+    (8, 1, [8]),
+    (6, 2, [3, 3]),
+    (17, 4, [4, 4, 4, 5]),
```
Oleksl888, line 21: does this data produce a result in equal parts?

app/test_split_integer.py
```diff
@@ -1,21 +1,67 @@
...
+        expected_result: list) -> None:
+    assert len(split_integer(value, number_of_parts)) == number_of_parts
+    assert max(expected_result) - min(expected_result) <= 1
 
 
-def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+@pytest.mark.parametrize("value, number_of_parts, expected_result", [
+    (8, 1, [8]),
+    (6, 2, [3, 3]),
```
Oleksl888, line 34: Test data should divide in one part - means second param must be 1 for this test

app/test_split_integer.py
```diff
@@ -1,21 +1,67 @@
...
+@pytest.mark.parametrize("value, number_of_parts", [
+    (8, 1),
+    (6, 2),
+    (17, 4),
+    (32, 6)
+])
+def test_should_add_zeros_when_value_is_less_than_number_of_parts(
+        value: int,
+        number_of_parts: int) -> None:
+    assert not value < number_of_parts
```
Oleksl888, line 67: How does this expression check that there are zeros in result?

Oleksl888 (changes requested):

PR #892, opened 2024-04-03
Solution 1

semermaks:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(2, 2) == [1, 1]
```
Oleksl888, line 5: How does this test check a sum?

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
```
Oleksl888, line 9: result data does not have equal parts

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(10, 1) == [10]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(15, 3) == [5, 5, 5]
```
Oleksl888, line 17: you should test if result is sorted on dataset that contains different numbers

Oleksl888 (changes requested):

PR #896, opened 2024-04-04
Develop

VitalyBashkiser:


Oleksl888 (changes requested):
You will use parametrizer in future tasks. Just follow the checklist and write 5 separate tests

LiudmylaKulinchenko (changes requested):
Just create 5 separate tests, like Oleksandr suggested

app/test_split_integer.py
```diff
@@ -1,21 +1,65 @@
+import unittest
 from app.split_integer import split_integer
 
 
-def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+class TestSplit(unittest.TestCase):
 
+    def test_split_integer_zero_parts(self) -> None:
```
Oleksl888, line 7: no need to delete original tests - you must use them as a boiler plate and add your code to the implementation.
Here you overcomplicated simple tests that should be one line.
No need to declare variables that will be used only once.
No need to write one-liner docstrings for tests - in this task optionally you can put an error message after assert to explain what went wrong e.g:
` assert split_integer(8, 1) == [8], value must be divided into a list with 1 element equal to value `

Oleksl888 (changes requested):

Oleksl888 (changes requested):
I will approve when you explain why did you rename the tests

VitalyBashkiser:
> I will approve when you explain why did you rename the tests

After the comment about not deleting old tests and after not realizing the previous error, I completely rewrote the tests, put new data and therefore new names for the functions.

Oleksl888:
> > I will approve when you explain why did you rename the tests
> 
> After the comment about not deleting old tests and after not realizing the previous error, I completely rewrote the tests, put new data and therefore new names for the functions.

You don't have to rename the tests here. You only need to write 1 simple assert statement under each existing test instead of a `pass` keyword

Oleksl888 (changes requested):
```
    Here you need to test that the sum of the result of the function equals initial value


    Here you must test data that splits equally for example 16 divided into 4 parts is 4,4,4,4


    If you split 16 by 1 it gives a list of one element - 16


    compare the result of the function with sorted version of the result (your data cannot be where all numbers are equal)


    you need to check if there will be zeros in the result

Oleksl888 (approved):
you got it!

PR #901, opened 2024-04-07
'Solution1'

yossosh:


Oleksl888 (changes requested):
doesnt pass the tests

Oleksl888 (approved):
the logic in  tests is a bit complicated but ok :)

PR #902, opened 2024-04-07
Solution

siaserb:


app/test_split_integer.py
```diff
@@ -2,20 +2,27 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(32, 6)) == 32, \
```
Oleksl888, line 5: We don't use '\' as line breakers in our projects. Please replace with parenthesis everywhere

Oleksl888 (changes requested):

Oleksl888 (approved):
Amazing

PR #903, opened 2024-04-07
Solution

Ruxxell54:


app/test_split_integer.py
```diff
@@ -2,20 +2,22 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(32, 6)
```
Oleksl888, line 5: no need to create extra variable if you not going to reuse this value

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Good job

PR #916, opened 2024-04-12
Solution

fishray33:


app/split_integer.py
```diff
@@ -5,3 +5,7 @@ def split_integer(value: int, number_of_parts: int) -> list:
         parts.append(value // parts_left)
         value -= next_number
     return parts
+
+
+if __name__ == "__main__":
+    print(f"{split_integer(2, 3)=}")
```
YuriiSmolii, line 11: You shouldn't add code here.

YuriiSmolii (changes requested):
Clear solution.

YuriiSmolii (approved):
LGTM

PR #922, opened 2024-04-15
Solution

sserkov84:


app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 22
+    parts = 4
+    result = split_integer(value, parts)
+    assert sum(result) == value
```
NazarSheremet, line 8: Based on clean code recommendations, will be better to do:

```suggestion
    assert sum(split_integer(22, 4)) == 22
```
or
```suggestion
    assert split_integer(22, 4) == [4, 4, 4, 6]
```
image

NazarSheremet (changes requested):
Need to implement suggestions for all tests.

NazarSheremet (approved):
Good job!

PR #924, opened 2024-04-15
Solution

dvkim28:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(22, 4)
+    assert sum(result) == 22
```
```suggestion
    assert sum(split_integer(22, 4)) == 22
```
no need to create variable, if it used only once

vsmutok (changes requested):

vasylhnatiuk (changes requested):
Знімок екрана 2024-04-16 о 17 48 29

app/test_split_integer.py
```diff
@@ -2,20 +2,22 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    split_integer(22, 4)
```
vasylhnatiuk, line 5: delete it as well

vasylhnatiuk (changes requested):

PR #927, opened 2024-04-19
'solution'

liudmylaperchevska:


app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(part for part in split_integer(32, 6)) == 32
```
vsmutok, line 5:
```suggestion
    assert sum(split_integer(32, 6)) == 32
```

app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
...
+    value = 6
+    number_of_parts = 2
+    if value / number_of_parts == value // number_of_parts:
+        assert len(set(split_integer(value, number_of_parts))) == 1
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    value = 8
+    assert split_integer(value, 1) == [value]
```
vsmutok, line 17:
```suggestion
    assert split_integer(8, 1) == [8]
```

no need to create variable if it use only once

vsmutok (changes requested):

liudmylaperchevska:
I'm not sure what should be simplified

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert len(set(split_integer(6, 2))) == 1
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert sum(split_integer(32, 6)) == 32
```
pavlopro, line 13: Please check the test name '...when_split_into_one_part' and make appropriate corrections.

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(17, 4) == sorted(split_integer(17, 4))
```
pavlopro, line 17: Instead of using sorted(split_integer(17, 4)), it would be better to hardcode the expected value

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

liudmylaperchevska:
I guess its better to say "finally, good job"

PR #929, opened 2024-04-23
Solution

oleg-gj:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    split = split_integer(value=20, number_of_parts=3)
```
pavlopro, line 5: There is no need to create a variable if you only use it once.

Please correct here and in methods below

pavlopro, line 5: Please write code according to the test name. The sum of the `split_integer `should be equal to the value.

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #952, opened 2024-05-23
added tests for split integer

AlexRozv:
solution

### DIFF
--- a/app/test_split_integer.py
+++ b/app/test_split_integer.py
@@ -2,20 +2,30 @@ from app.split_integer import split_integer
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (
+        sum(split_integer(32, 6)) == 32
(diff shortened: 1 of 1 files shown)


app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
...
+    assert (
+        len(set(split_integer(36, 6))) == 1
+    ), "Split parts should be equal when value divisible by parts"
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert (
+        split_integer(32, 1)[0] == 32
+    ), "Split part should be equal to initial value if parts == 1"
```
LiudmylaKulinchenko, line 19: But what if there are more than 1 value in a result list? Your test should cover this case. Just compare functioin result to an expected list

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌷

PR #957, opened 2024-05-27
Solution

SuskyiVolodymyr:


Oleg-s-s (approved):
Idk how to feel about creating these temp values everywhere, but overall it ok

PR #979, opened 2024-06-06
solution

NikitaNikolaitsev:


app/test_split_integer.py
```diff
@@ -2,20 +2,22 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(8, 2)) == 8
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(8, 2) == [4, 4]
```
DevSpaciX, line 9: Add an error message. It's good when the user understands what the problem is

app/test_split_integer.py
```diff
@@ -2,20 +2,22 @@
...
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(17, 4) == [4, 4, 4, 5]
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(1, 2) == [0, 1]
+# Do I specifically need to break one of the tests?
+# If I break it, then pytest app/ won't pass
```
DevSpaciX, line 23: The test failed because of pass param . Try to add error messages as I say below , and create another PR

DevSpaciX (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
...
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(17, 4) == [4, 4, 4, 5], \
+        "Parts should be sorted when they are not equal"
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(1, 2) == [0, 1], \
```
DevSpaciX, line 25: Remove this breaking lines ( \ ) . Try to use more beautiful variants

DevSpaciX (changes requested):

app/test_split_integer.py
NikitaNikolaitsev: ok only this problem?

DevSpaciX (approved):
Great !

DevSpaciX (changes requested):
Make all test passed

PR #980, opened 2024-06-06
'Solution'

dsokolovskiy:


app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 32
+    number_of_parts = 6
+    assert sum(split_integer(value, number_of_parts)) == value
```
DevSpaciX, line 7: Add an error message. It's good when the user understands what the problem is

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #987, opened 2024-06-06
'Solution'

ghost:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(32, 6)) == 32
```
DevSpaciX, line 5: Add an error message. It's good when the user understands what the problem is

DevSpaciX (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(23, 5) == [4, 4, 5, 5, 5]
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(1, 2) == [0, 1]
```
DevSpaciX, line 21: Why you did not add error messages here too ?

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #988, opened 2024-06-06
Develop

Irina17191:


app/test_split_integer.py
```diff
@@ -2,20 +2,33 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(8, 3)) == 8
+    assert sum(split_integer(12, 4)) == 12
+    assert sum(split_integer(20, 5)) == 20
```
DevSpaciX, line 7: Add an error message. It's good when the user understands what the problem is

DevSpaciX (changes requested):

app/test_split_integer.py
Irina17191: Hello) I added messages and rewrote tests using mark.parametrize

DevSpaciX (approved):
Nice!

PR #991, opened 2024-06-07
The solution of the task

Ja0ni:


Dimosphen1 (changes requested):
Make sure that all checks pass (mentor tests are failing)

app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = split_integer(3, 3)
```
pavlopro, line 5: There are no need to have this 'value' variable 
Simplify this to `sum(split_integer(3, 3)) == 3`

Fix it in all other places

pavlopro (changes requested):

vsmutok (approved):
Good Job!

PR #998, opened 2024-06-12
Solution

MaksymProtsak:


vasylhnatiuk (changes requested):
read checklist and fix your code https://github.com/mate-academy/py-split-integer/blob/master/checklist.md

app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (sum(split_integer(6, 2)) == 6),\
+        "Sum of the parts should be equal to_value."
```
vsmutok, line 6: better to use `()` then `\`
please fix everywhere

vsmutok (changes requested):

app/test_split_integer.py
MaksymProtsak: Like that?
`def test_sum_of_the_parts_should_be_equal_to_value() -> None:
    assert (sum(split_integer(6, 2)) == 6), (
            "Sum of the parts should be equal to_value.")`

MaksymProtsak: Done.

vsmutok (approved):
Great!

PR #1009, opened 2024-06-25
Solution

EugeneDolinskiy:


EugeneDolinskiy:
It is right? Or did you mean that instead of creating variables, i can just insert a function call?

EugeneDolinskiy:
I get it)
Sorry for bothering you by my pushes!

PR #1011, opened 2024-06-26
+

AndriZhok:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (split_integer(8, 1) == [8])
```
pavlopro, line 5: No need for those parentheses
Also fix in methods below as well

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #1018, opened 2024-07-02
Try Solution

D0mln0:


app/test_split_integer.py
```diff
@@ -1,21 +1,52 @@
...
+        (17, 4, [4, 4, 4, 5]),
+        (32, 6, [5, 5, 5, 5, 6, 6]),
+    ]
+)
+def test_parts_should_be_sorted_when_they_are_not_equal(
+    value: int, number_of_parts: int, result: list
+) -> None:
+    func = split_integer(value, number_of_parts)
+    expected = sorted(func)
+    assert (func == expected), \
```
vasylhnatiuk, line 46: try to avoid using \

app/test_split_integer.py
```diff
@@ -1,21 +1,52 @@
...
+        f"{split_integer(6, 2)} should be equal parts"
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
-
-
-def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert (split_integer(8, 1) == [8]), \
```
vasylhnatiuk, line 28: try to avoid using \

vasylhnatiuk: do it everywhere

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
.

D0mln0:
why it is not a good idea to use '\'?

vasylhnatiuk (approved):
because it looks better with ()

PR #1040, opened 2024-07-22
Implement tests

anton-bilous:


app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
...
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(100, 1) == [100]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    split_result = split_integer(32, 6)
+    assert split_result == sorted(split_result)
```
LiudmylaKulinchenko, line 19: Rewrite it in one line, like you do in other tests

app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert set(split_integer(16, 4)) == {4}
```
LiudmylaKulinchenko, line 9: This solution is correct, I'm just thinking will it be as descriptive as writing the result directly? 🤔 What do you think?

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌞

PR #1042, opened 2024-07-23
"Solution py-split-integer"

Olga-Gynguliak:


app/test_split_integer.py
```diff
@@ -2,20 +2,27 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(8, 1)) == 8
 
 
-def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+def \
```
vasylhnatiuk, line 8: why you need \ here?? delete it everywhere

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
fix flake8

PR #1043, opened 2024-07-23
Implemented test_split_integer.py

OleksiukStepan:


LiudmylaKulinchenko (changes requested):
In this case logic is pretty simple and adding several values for the same test doesn't actually helpfull. It could be helpfull if we would need to test negative or large numbers. In this case (6, 2) or (17, 4) will test the same functionality

PR #1044, opened 2024-07-23
Implemented solution

dimak20:


app/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
...
+        split_integer(76, 55)
+        == sorted(split_integer(76, 55))
+    )
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert (
+        min(split_integer(25, 60)) == 0
+    )
```
LiudmylaKulinchenko, line 32: How do we know how many zeros will be in the result and if they placed correctly with this test? Let's fix it 💪

app/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
...
+    assert (
+        sum(split_integer(100, 7)) == 100
+    )
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert (
+        max(split_integer(555, 555)) == 1
+    )
```
LiudmylaKulinchenko, line 13: same here. How can we be sure that there are equal amount of parts? Also you should test here number, that is devisible by the next number, like 5 and 25. Not the same numbers

LiudmylaKulinchenko (changes requested):
Let's also don't split into lines if it fits in one, in this case logic is pretty simple, so we can do that!

app/test_split_integer.py
```diff
@@ -2,20 +2,32 @@
...
+    assert (
+        split_integer(5, 5) == [1, 1, 1, 1, 1]
+    )
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert (
+        split_integer(95, 1) == [95]
+    )
```
MNDonut, line 19: doesn't it fit in 1 line? fix everywhere

MNDonut (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
```

oliasymak24 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
...
-    pass
+    assert (
+        sum(split_integer(100, 7)) == 100
+    ), "sum of parts must be equal to value"
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert (
+        split_integer(5, 5) == [1, 1, 1, 1, 1]
```
Oleksl888, line 12: this test and the message suggests that we are testing when values is equal to number of parts , but actually we are testing if result will contain equal parts if value is divisible without a remainder

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1050, opened 2024-07-30
Solution

DSahalatyi:


app/test_split_integer.py
```diff
@@ -2,20 +2,29 @@
...
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(8, 1) == [8]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    result = split_integer(17, 4)
+    assert result == sorted(result)
```
LiudmylaKulinchenko, line 22: Read checklist. Make sure to fix everywhere throughout the code

app/test_split_integer.py
```diff
@@ -2,20 +2,29 @@
...
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    result = split_integer(8, 12)
+    assert result[0:4] == [0, 0, 0, 0]
```
LiudmylaKulinchenko, line 27: You don't test numbers after 4th index, how are you sure they're correct?

app/test_split_integer.py
```diff
@@ -2,20 +2,29 @@
...
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    result = split_integer(8, 2)
+    assert result[0] == result[1]
```
LiudmylaKulinchenko, line 10: It's ok, but lets rewrite in one line and remove variable to make it more self-explanatory

LiudmylaKulinchenko (changes requested):

app/test_split_integer.py
```diff
@@ -1,21 +1,39 @@
 from app.split_integer import split_integer
```

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1051, opened 2024-08-01
create five test function

MaksimGolev:


app/test_split_integer.py
```diff
@@ -2,20 +2,36 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 17
+    number_of_parts = 4
+    result = split_integer(value, number_of_parts)
+    assert sum(result) == value
```
NazarSheremet, line 8: Based on clean code recommendations, will be better to do:

```suggestion
    assert sum(split_integer(17, 4)) == 17
```
image

NazarSheremet (changes requested):
Need to reformat code

YuriiSmolii (approved):
LGTM

PR #1054, opened 2024-08-04
solution

HalynaRohatska:


app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
...
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    result = split_integer(32, 6)
+    assert result == [5, 5, 5, 5, 6, 6]
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    result = split_integer(0, 4)
+    assert result == [0, 0, 0, 0]
```
LiudmylaKulinchenko, line 25: Let's test bigger number to test as well if zeros are placed in correct places in the result list

app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
...
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(8, 1) == [8]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    result = split_integer(32, 6)
+    assert result == [5, 5, 5, 5, 6, 6]
```
LiudmylaKulinchenko, line 20: Read checklist. Make sure to fix everywhere throughout the code

app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
...
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = sum(split_integer(17, 4))
+    assert result == 17
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    result = split_integer(6, 2)
+    assert result[0] == result[1]
```
LiudmylaKulinchenko, line 11: It's ok, but lets rewrite in one line and remove variable to make it more self-explanatory

LiudmylaKulinchenko (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
```

oliasymak24 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,29 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17, \
```
Oleksl888, line 5: We don't use `\ `when breaking lines as per style convention in our projects - use parenthesis instead:
```
assert (

Oleksl888 (changes requested):

PR #1055, opened 2024-08-05
Solution

dartomOOv:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17
```
oliasymak24, line 5: Add an error message. It's good when the user understands what the problem is

oliasymak24 (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🌞

PR #1057, opened 2024-08-05
writing first test solution

Narberal90:


app/test_split_integer.py
```diff
@@ -1,21 +1,23 @@
+
+
 from app.split_integer import split_integer
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert len(split_integer(32, 6)) == 6
```
LiudmylaKulinchenko, line 7: sum of the parts should be equal to value -- you should summurize parts here, not count them

LiudmylaKulinchenko (changes requested):

MNDonut (approved):

PR #1063, opened 2024-08-07
Task Solution

achalaidiuk:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
```

oliasymak24 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(5, 1) == [5], \
```
Oleksl888, line 5: How does this test check for sum of parts?

app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
...
+        "(5, 1) should return [5]"
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(36, 6) == [6, 6, 6, 6, 6, 6], \
```
Oleksl888, line 10: We don't use \ when breaking lines as per style convention in our projects - use parenthesis instead:
```
assert (

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1065, opened 2024-08-07
Solution

dpiuro:


app/test_split_integer.py
```diff
@@ -2,20 +2,26 @@
```
oliasymak24, line 3: Add an error message for each task. It's good when the user understands what the problem is

app/split_integer.py
```diff
@@ -1,7 +1,5 @@
...
-    parts = []
-    for parts_left in range(number_of_parts, 0, -1):
-        next_number = value // parts_left
-        parts.append(value // parts_left)
-        value -= next_number
-    return parts
+    parts = [value // number_of_parts] * number_of_parts
+    for i in range(value % number_of_parts):
+        parts[i] += 1
+    return sorted(parts)
```
oliasymak24, line 5: Ypu don't need to change this function

oliasymak24 (changes requested):

Oleksl888 (approved):
Approved. Though it would make sense to use mark.parametrise here and not make multiple assert statements

PR #1066, opened 2024-08-07
'Solution'

ABaidin:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
```
oliasymak24, line 3: Add an error message for each task. It's good when the user understands what the problem is

oliasymak24 (changes requested):

ABaidin:
Tests already have descriptive names, so I tried to not repeat them in error messages

app/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (
+        sum(split_integer(32, 6)) == 32
+    ), "5 + 5 + 5 + 5 + 6 + 6 should be equal to 32"
```
Oleksl888, line 7: Suggesting to work on your exception messages. Here `5 + 5 + 5 + 5 + 6 + 6`doesn't relate to the assert statement at all.
You could do something like this:
`The sum of the parts should be equal to the original value`

app/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert (
+        split_integer(25, 5) == [5, 5, 5, 5, 5]
+    ), "25 is divisible by 5 so it should return [5, 5, 5, 5, 5]"
```
Oleksl888, line 13: No need to explain that 25 is divisible by 5.
`List should have equal parts if the original value is divisible by the number of parts`

app/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert (
+        split_integer(10, 1) == [10]
+    ), ("'numbers_of_parts' = 1, so func should return "
+        "list with one element equal to 'value'")
```
Oleksl888, line 20: If the number of parts is 1, the resulting list should contain the original value

app/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
...
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert (
+        split_integer(18, 4) == sorted((split_integer(18, 4)))
+    ), "You should sort your resulting list"
```
Oleksl888, line 27: Resulting list should be sorted when parts are not equal

app/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
...
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert (
+        split_integer(2, 4) == [0, 0, 1, 1]
+    ), "2 < 4, so you need to add zeros to match number of parts"
```
Oleksl888, line 32: If the value is less than the number of parts , zeros must be added to match the number of parts.

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Good job 💪

PR #1067, opened 2024-08-07
Solution Politov

Alexandr-Politov:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert set(split_integer(36, 6)) == {6}
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert len(split_integer(36, 1)) == 1
```
Oleksl888, line 13: this test checks that resulting list will have only one element but it doesn't check that this element will be equal to initial value

Oleksl888 (changes requested):

oliasymak24 (approved):
Good job!

PR #1068, opened 2024-08-07
Solution of py-split-integer

AndriiHamasa:


app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    error_mes = "Sum of the parts should be equal to `value`"
```
Oleksl888, line 5: No need to store this in a variable - add exception message after a comma, break lines using parenthesis so you don't have flake errors

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1069, opened 2024-08-07
commit

panichqa:


app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
...
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(8, 1) == [8]
+    assert split_integer(15, 1) == [15]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    result = split_integer(17, 4)
```
oliasymak24, line 19: No need to create separate variable here

app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
...
+    assert result == sorted(result)
+    assert result == [4, 4, 4, 5]
```
oliasymak24, line 21: You can do it in one line

app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
```
oliasymak24, line 3: Add an error message for each test . It's good when the user understands what the problem is

oliasymak24 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,36 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(8, 1)) == 8,\
```
Oleksl888, line 5: We don't use `\` when breaking lines as per style convention in our projects - use parenthesis instead:
```
assert (

File: app/test_split_integer.py [hunk 5]
```diff
@@ -2,20 +2,36 @@
...
+        "Sum of the parts should be equal to the original value."
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(8, 4) == [2, 2, 2, 2],\
```
Oleksl888, line 10: If you use different data sets on the same test it would make sense to parametrize them to avoid code repetition. 
You can optionally use mark,parametrize if you want to be a :star: (it is not required by task definition)

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1071, opened 2024-08-07
Solution

Bogdan-Rylov:


app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17
```
Oleksl888, line 5: Please provide an exception message in each test

Oleksl888 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,26 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17, \
+        "The sum of the parts should be equal to the value"
```
Oleksl888, line 6: We don't use `\ `when breaking lines as per style convention in our projects - use parenthesis instead:
```
assert (

File: app/test_split_integer.py [hunk 3]
```diff
@@ -2,20 +2,26 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(25, 5) == [5, 5, 5, 5, 5], \
+        "The result length should be equal to the number of parts"
```
Oleksl888, line 11: We are not really testing for length - more appropriate would be to say that all elements in result should be equal to each other

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1072, opened 2024-08-08
Solution

olena4reunited:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
```

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(6, 2) == [3, 3]
```
oliasymak24, line 9: Usually better to put for test more complicated case, for example `split_integer(36, 6) == [6, 6, 6, 6, 6, 6]`

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1073, opened 2024-08-08
first

egor250330:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
```

app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(32, 6)) == 32, \
```
Oleksl888, line 5: We don't use `\` when breaking lines as per style convention in our projects - use parenthesis instead:
```
assert (

Oleksl888 (changes requested):

PR #1075, opened 2024-08-08
task done

Sergi0bbb:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(8, 1) == [8]
```
Oleksl888, line 5: How does this check that sum of result equals to value?

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(6, 2) == [3, 3]
```
Oleksl888, line 9: Please provide an exception message in each test

Oleksl888 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,32 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(9, 2) == [4, 5], (
```
oliasymak24, line 5: Where are you calculatinh the sum?

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1076, opened 2024-08-08
Resolve

ihorutkin:


app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17
```
Oleksl888, line 5: Please provide an exception message in each test

app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(6, 2) == [3, 3]
+    assert split_integer(17, 4) == [4, 4, 4, 5]
```
Oleksl888, line 10: this assertion is correct but input data is wrong - result is not splitting in equal parts

Oleksl888 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17, \
+        "test sum of the parts should be equal to value"
```
LiudmylaKulinchenko, line 6: Try not to use `\` as a line breaker, you can use parentheses instead

LiudmylaKulinchenko (changes requested):

app/test_split_integer.py
ihorutkin: Hi, when I use parentheses i get this type of error: assertion is always true, perhaps remove parentheses?

oliasymak24 (approved):

PR #1077, opened 2024-08-08
Solution

mnezbrytska:


app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (
+        split_integer(15, 3) == [5, 5, 5]
```
Oleksl888, line 6: The name of the test assumes you will calculate sum of the result

Oleksl888 (changes requested):

app/test_split_integer.py
mnezbrytska: I don't understand what I need to change, the sum of all items in the list should be equal value.

Oleksl888: The name of the test `def test_sum_of_the_parts_should_be_equal_to_value() -> None:` assumes that we need to check that SUM of the parts should be equal to value. It suggests that in the test we must calculate sum of the resulting list and compare it vs original value.
No need to rename this test function - just change the logic of a test slightly

Oleksl888 (changes requested):

app/test_split_integer.py
```diff
@@ -1,21 +1,33 @@
 from app.split_integer import split_integer
 
 
-def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+def test_all_element_in_list_should_be_equal_to_value() -> None:
+    result = split_integer(15, 3)
+    expected_sum = 15
```
LiudmylaKulinchenko, line 6: No need to create variables here

LiudmylaKulinchenko (changes requested):

oliasymak24 (approved):

PR #1078, opened 2024-08-08
implement py-split-integer

Maksym-Turenko:


app/test_split_integer.py
```diff
@@ -1,21 +1,49 @@
+# test_split_integer.py
 from app.split_integer import split_integer
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 20
```
Oleksl888, line 6: no need to create variables that you don't reuse - these are simple tests and should be done pretty much in a single line of code

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1080, opened 2024-08-11
solving

Poznanskyi:


app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
...
-    pass
+    assert (
+        split_integer(6, 2) == [3, 3]
+    ), "Should split into equal parts when value divisible by parts"
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert (
+        split_integer(17, 4) == [4, 4, 4, 5]
```
Oleksl888, line 18: This data doesn't test that value is split in one part

Oleksl888 (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,30 @@
...
+    ), "Should split into 4 parts"
```
LiudmylaKulinchenko, line 19: Your test name isnt's similar to your message. You should've changed test logic after Oleksandrs comment, not message

LiudmylaKulinchenko (changes requested):

PR #1098, opened 2024-08-27
add solution

MRozdoba:


app/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert (
```
pavlopro, line 5: No need to have those brackets
Please fix in places below as well

app/test_split_integer.py
```diff
@@ -2,20 +2,31 @@
...
+        sum(split_integer(32, 6)) == 32
+    )
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    for element in split_integer(30, 5):
+        assert (
+            element == 6
+        )
```
pavlopro, line 14: This is not correct

pavlopro (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #1099, opened 2024-08-27
Solution

ip-ilya:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    split_integer_result = split_integer(17, 4)
```
pavlopro, line 5: No need to have this variable
Directly compare `sum(split_integer(17, 4)) == 17`

Fix in places below

pavlopro (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
...
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(17, 4) == sorted(split_integer(17, 4))
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    split_integer_result = split_integer(3, 5)
+    assert split_integer_result == [0, 0, 1, 1, 1]
```
vsmutok, line 22:
```suggestion
    assert split_integer(3, 5) == [0, 0, 1, 1, 1]
```

vsmutok (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(17, 4)) == 17
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert len(set(split_integer(8, 4))) == 1
```
Dimosphen1, line 9: Sometimes it is preferable to make an explicit comparison instead of additional modifications over data to improve the readability
```suggestion
    assert split_integer(8, 4) == [2, 2, 2, 2]
```

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #1106, opened 2024-09-04
split integer

AndreyHMkr:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(6, 2) == [3, 3]
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(17, 4) == [4, 4, 4, 5]
```
pavlopro, line 13: This is not correct realization

...split into one part

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(8, 1)) == 8
```
pavlopro, line 5: This is not correct realization
it means sum of the parts, not one part

pavlopro (changes requested):

LLkaia (approved):

PR #1112, opened 2024-09-09
Solution, tests passed, flake8 - ok

samuel-edmund-morgan:


NazarSheremet (changes requested):
Need to update your solution according to task checklist recomendation 
image

NazarSheremet (approved):
Good job!

PR #1120, opened 2024-09-19
solution for split integer

dissom:


vasylhnatiuk (changes requested):
https://github.com/mate-academy/py-split-integer/blob/master/checklist.md

read check list and fix your solution

app/test_split_integer.py
```diff
@@ -2,20 +2,46 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value = 15
```
vasylhnatiuk, line 5: you still use useless varibles

vasylhnatiuk (changes requested):

PR #1136, opened 2024-10-04
'Solution'

Alona3:


app/test_split_integer.py
```diff
@@ -2,20 +2,35 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    for value, number_of_parts in [(8, 1), (6, 2), (17, 4), (32, 6)]:
```
pavlopro, line 5: No need to have those iteration of tested values. Simplify each tests
for example this test you can simplify to `assert 6 == sum(split_integer(6, 2))`

pavlopro (changes requested):

pavlopro (approved):

PR #1145, opened 2024-10-12
split-integer solution

94nj111:


app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    goal = split_integer(16, 4)
+    assert sum(goal) == 16
```
AnyoneClown, line 6: We can refactor it in single line. Fix in all cases.

```suggestion
    assert sum(split_integer(16, 4)) == 16
```

app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
...
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    goal = split_integer(6, 2)
+    assert goal[0] == goal[1]
```
AnyoneClown, line 11: Add an error message(for all tests). It's good when the user understands what the problem is

AnyoneClown (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,26 @@
...
-    pass
+    assert (sum(split_integer(16, 4)) == 16),\
+        "sum of the parts should be equal to value"
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    goal = split_integer(6, 2)
+    assert goal[0] == goal[1],\
+        "should split into equal parts when value divisible by parts"
```
AnyoneClown, line 12: Don't use `\`. Fix it everywhere

```suggestion
    assert goal[0] == goal[1], (
        "should split into equal parts when value divisible by parts"
    )
```

AnyoneClown (changes requested):

PR #1150, opened 2024-10-16
Solution

sofiiasavkova:


app/split_integer.py
```diff
@@ -1,4 +1,7 @@
 def split_integer(value: int, number_of_parts: int) -> list:
+    if value < number_of_parts:
+        return [1] * value + [0] * (number_of_parts - value)
```
Mxbely, line 3: I think you don't need add this validate value, because you have a condition. Remove it
image

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(8, 1) == [8]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(17, 4) == sorted(split_integer(17, 4))
```
Mxbely, line 17: this test don't check expected results, if you has been broken this function, you can pass this test, but you will not get expected results

app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(3, 5) == [1, 1, 1, 0, 0]
```
Mxbely, line 21: And if you don't need any validation, need to change this line (21)
Hint: you can use result of this function

Mxbely (changes requested):

PR #1162, opened 2024-10-17
solution

dxrrkwm:


Bondzik-S (changes requested):
how about assertion messages for each test?)

PR #1175, opened 2024-10-20
solution

YaYaYaroslav:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
...
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(7, 1) == [7]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    assert split_integer(17, 4) == sorted(split_integer(17, 4))
```
Arsen-hrynevych, line 17: it’s not ideal to call split_integer() twice. A cleaner approach would be to call it once and then check for sorting.

Arsen-hrynevych (approved):
Ok ;)

PR #1233, opened 2024-12-04
Solution_v1.0

Rozdorozhnii:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    goals = split_integer(32, 6)
+    assert sum(goals) == 32
```
Nattalli, line 6: All functions can be written in one line, it is more readable

Nattalli: Add an error message(for all tests). It's good when the user understands what the problem is

Nattalli (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,26 @@
...
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(8, 1) == [8], \
+        "Should return part equals to value when split into one part"
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    goals = split_integer(17, 4)
+    assert goals[-1] > goals[0], \
```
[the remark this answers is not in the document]
Rozdorozhnii: @Alkaponees I made what you ask but as for me solution become more complicated now

Nattalli (approved):

PR #1256, opened 2024-12-18
error

maxakim0v:


app/test_split_integer.py
```diff
@@ -1,21 +1,129 @@
...
+)
+def test_max_and_min_difference_should_be_less_that_one(
+    value: int,
+    number_of_parts: int
+) -> None:
+    result = split_integer(value, number_of_parts)
+    assert max(result) - min(result) <= 1, (
+        "Difference of max and min number in the array should be <= 1"
+    )
+
```
petrykivd, line 25: Please read checklist before pushing your solution!
Don't create unnecessary variables for simple tests:

petrykivd: FIx it in all places

app/test_split_integer.py
```diff
@@ -1,21 +1,129 @@
...
+@pytest.mark.parametrize(
+    "value,number_of_parts",
+    [
+        (1, 2),
+        (1, 10),
+        (8, 9),
+        (20, 25),
+        (100, 102)
+    ]
+)
```
petrykivd, line 120: Are you sure you need to use parameterization here?

petrykivd (changes requested):

PR #1259, opened 2024-12-18
Solution

bogdAAAn1:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(32, 6)
```
Chukhno-Maksym, line 5: You should compare the sums of the lists (sum()), not the exact contents or order of the elements.

Chukhno-Maksym (approved):

PR #1278, opened 2024-12-20
solution_v.1.0

yelizaveta-zh:


L1nk3rrr (changes requested):
At task, you need to use `pytest`, so it looks like `mark.parametrize` can be helpful here!

L1nk3rrr (changes requested):
actually my comment was to do parametrize for all functions, not made from 4 -> 2
also it will help you to add a messages when assertation fails

PR #1279, opened 2024-12-21
Solution

bodiakof:


L1nk3rrr (changes requested):
Well done!
Here are some styling points.

1. Try to stick for one style with parameterized, because now you have 3 different styles for parameterized input and it's hard to read on some moments.
2. When you assert write a messages after assertations, that will improve tests understanding and what is going wrong

L1nk3rrr (approved):
Well done!

PR #1314, opened 2025-01-16
done

Leo9siy:


app/test_split_integer.py
```diff
@@ -2,20 +2,34 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(8, 1)
+    assert sum(result) == 8
```
vsmutok: According to the checklist, avoid creating unnecessary variables for simple tests. You can directly assert the function call: `assert sum(split_integer(8, 1)) == 8`

vsmutok (changes requested):

app/test_split_integer.py
```diff
@@ -2,20 +2,27 @@
...
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    result = split_integer(8, 2)
+    assert result[0] == result[1]
 
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    result = split_integer(8, 1)
+    assert result[0] == 8
```
vsmutok, line 18: Consider removing the `result` variable and directly asserting: `assert split_integer(8, 1)[0] == 8.`

vsmutok (changes requested):

PR #1343, opened 2025-02-04
Complete py tests to function split_integer

kram3ko:


Baranotik15 (approved):
Everything is correct, but I’m not sure that it is necessary to complicate the conditions in the tests and not clearly indicate the correct answer, as for example

    assert split_integer(3, 8).count(0) == 8 - 3
    
I think better something like this:

    assert split_integer(3, 8) == [0, 0, 0, 0, 0, 1, 1, 1]
    
but maybe I'm wrong

PR #1344, opened 2025-02-05
Solution

Baranotik15:


app/test_split_integer.py
```diff
@@ -2,20 +2,33 @@
...
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    for value, parts in [(10, 3), (7, 3), (11, 5), (25, 7)]:
+        assert sum(split_integer(value, parts)) == value
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert split_integer(10, 5) == [2, 2, 2, 2, 2]
```
kram3ko, line 10: it gives guarantee that always spleet on same parts? currently in this case it do.
if split_integer(101, 100) == [1..............................2] maybe set ? or something?

app/test_split_integer.py
```diff
@@ -2,20 +2,33 @@
...
+        result = split_integer(value, parts)
+
+    assert result == sorted(result), f"Result {result} is not sorted"
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(1, 5) == [0, 0, 0, 0, 1]
+    assert split_integer(0, 5) == [0, 0, 0, 0, 0]
+
```
kram3ko, line 27: maybe check with count diference between  values?.

kram3ko (commented):
try to think.

kram3ko (approved):
Good job :) maybe i'm not right? ;)

PR #1390, opened 2025-03-03
'Solution'

AndyStarGitHub:


app/test_split_integer.py
```diff
@@ -2,20 +2,69 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
```
Oneyura, line 4: In this task i we need to check if sum of the parts equal to value input.
It can be realise like you done but if AI dont approve, just think about this.

Oneyura (approved):
well done

PR #1396, opened 2025-03-05
well done

BronkstoneBro:


app/test_split_integer.py
```diff
@@ -2,20 +2,21 @@
...
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    actual = split_integer(17, 4)
+    assert actual == sorted(actual)
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
```
PrimeGlorious, line 22: Value must be less than number of parts. Test should return list with zeros

PrimeGlorious (changes requested):

PR #1410, opened 2025-03-07
Solution

Lerity12:


app/test_split_integer.py
```diff
@@ -1,21 +1,27 @@
...
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    result = split_integer(3, 5)
+    assert result == [0, 0, 1, 1, 1]
+
+
+if __name__ == "__main__":
+    pytest.main()
```
Boyok1337, line 27: I guess, if we run tests via command, so we don`t need this construction here

Lerity12: Thanks for your comment. Agree, it is unnecessary, I used it only to bypass flake8 error (it didn't seem to affect the tests in any way, so I decided to leave it)

oliasymak24 (changes requested):

PR #1416, opened 2025-03-09
Solution Split integer

Roksolana-K:


app/test_split_integer.py
oliasymak24, line 1: Add an error message. It's good when the user understands what the problem is

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1476, opened 2025-04-13
Solution

4zik88:


app/test_split_integer.py
vsmutok, line 1: In this task, there should be only one test for each test function.

vsmutok (changes requested):

app/test_split_integer.py
4zik88: FIxed

app/test_split_integer.py
```diff
@@ -2,20 +2,28 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    result = split_integer(17, 4)
+    assert sum(result) == 17
```
vsmutok: image

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1574, opened 2025-07-09
my-solution

Orixara:


Artcrafterrra (approved):
Good solution according to @pytest.mark.parametrize practice, but I think it’s not equal to the task. Also, we don’t need to change split_integer.py at all.

PR #1585, opened 2025-07-11
solution

Sashoxxx:


app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(18, 2)) == 18
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    result = split_integer(16, 4)
```
MakovM: Don't create unnecessary variables for simple tests
It's better to specify the result we expect in assert, for example:
assert split_integer(16, 4) == [4, 4, 4, 4]

app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
...
 
 def test_should_return_part_equals_to_value_when_split_into_one_part() -> None:
-    pass
+    assert split_integer(10, 1) == [10]
+    assert split_integer(121, 1) == [121]
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    result = split_integer(7, 3)
```
MakovM, line 18: You don't need to use "sorted", it is better to immediately indicate what result is expected
For example:

app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
...
+    assert result == sorted(result)
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    assert split_integer(2, 4) == [0, 0, 1, 1]
```
MakovM, line 23: No need to use two checks that perform the same functionality

MakovM (changes requested):

app/test_split_integer.py
Sashoxxx: > No need to use two checks that perform the same functionality
These two asserts test the same behavior: the function should work correctly when value < parts.
They do not test different functions or scenarios, but rather different inputs for the same situation.

app/test_split_integer.py
Sashoxxx: This test is not about checking the exact output values, but about ensuring a general property of the result — that the parts are sorted in non-decreasing order.
That’s why assert result == sorted(result) is appropriate here.

MakovM (approved):

PR #1589, opened 2025-07-11
Solution

SensibleN00B:


Salmon-eat (approved):
ok, but the Readme condition is not met, I don't see it. Not imported import pytest @pytest.mark.parametrize(...).

PR #1600, opened 2025-07-14
Solution

reinvvay:


trellle (approved):
you can also implement that by @pytest.mark.parametrize

PR #1612, opened 2025-07-23
Solution

Evgeniya0208:


app/test_split_integer.py
```diff
@@ -2,20 +2,25 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    parts = split_integer(32, 6)
+    assert sum(parts) == 32
```
vsmutok:
```suggestion
    assert sum(split_integer(32, 6)) == 32
```
Don't create unnecessary variables for simple tests

vsmutok (changes requested):

app/test_split_integer.py
Evgeniya0208: fixed

app/test_split_integer.py
```diff
@@ -2,20 +2,24 @@
...
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(32, 6)) == 32
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    parts = split_integer(6, 2)
+    assert parts == [3, 3]
```
vsmutok: Don't create unnecessary variables for simple tests

vsmutok (changes requested):

app/test_split_integer.py
Evgeniya0208: fixed

vsmutok (approved):
Great!

PR #1662, opened 2025-09-09
Solution #1

VasylAzarov:


app/test_split_integer.py
```diff
@@ -2,20 +2,26 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    value, parts = 17, 4
+    result = split_integer(value, parts)
+    assert sum(result) == value
```
vsmutok, line 7: The test checks the sum of parts but does not assert that the returned list has exactly number_of_parts elements. The description requires the function to "return an array containing exactly number_of_parts integer elements:" — add an assertion such as assert len(split_integer(17, 4)) == 4 (and similar assertions for other examples) to satisfy this requirement .

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1731, opened 2025-10-30
'Solotion'

Vorongor:


vadym-zahrebelnyi (approved):
Good job! But I'm not sure whether we should add messages to the assert statements or not.

PR #1749, opened 2025-11-06
Added solution

vadym-zahrebelnyi:


app/test_split_integer.py
```diff
@@ -2,20 +2,36 @@
...
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert sum(split_integer(32, 6)) == 32, (
+        "The summa of the parts should be equal to `value`"
+    )
 
 
 def test_should_split_into_equal_parts_when_value_divisible_by_parts() -> None:
-    pass
+    assert len(set(split_integer(24, 6))) == 1, (
```
leemonch1k-space: To be sure in func return value, better check the output value, then length, but it's an interesting opinion.

app/test_split_integer.py
```diff
@@ -2,20 +2,36 @@
...
-    pass
+    assert split_integer(8, 1) == [8], (
+        "The function return part equal to `value` "
+        "when `number_of_parts` is `1`"
+    )
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    result = split_integer(17, 4)
```
leemonch1k-space: according to checklist, we shouldn't create variables if code won't be too long

leemonch1k-space (changes requested):
Good job, all works correctly, but rewrite the function in 24's row, without variable, to complete checklist requirements

PR #1758, opened 2025-11-10
completed task

ostapshpeha:


app/test_split_integer.py
```diff
@@ -1,21 +1,48 @@
...
 
 
 def test_parts_should_be_sorted_when_they_are_not_equal() -> None:
-    pass
+    result = split_integer(17, 4)
+
+    assert len(result) == 4
+    assert result == sorted(result)
+    assert max(result) - min(result) <= 1
+    assert max(result) != min(result)
```
vakt159: I think there are too many useless assetion: you can just: 
assert split_integer(17,4) == [2, 5, 5, 5] 
and it has to fulfill test

app/test_split_integer.py
```diff
@@ -1,21 +1,48 @@
...
 
 
 def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+    result = split_integer(4, 6)
+
+    assert len(result) == 6
+    assert sum(result) == 4
+    assert result == sorted(result)
+    assert result.count(0) == 2 and result.count(1) == 4
```
vakt159: Same thing here:
assert split_integer(4, 6) == [0, 0, 1, 1, 1, 1]

vakt159 (changes requested):
Simplify your tests and everything have to be good

PR #1761, opened 2025-11-11
Solution

Guillelmina8:


app/split_integer.py
```diff
@@ -1,5 +1,8 @@
 def split_integer(value: int, number_of_parts: int) -> list:
     parts = []
+    if value < number_of_parts:
+        temporary = str(value) + "0"
+        value = int(temporary)
```
Iventyk: I think you shouldn't change the original function, only write tests

app/test_split_integer.py
```diff
@@ -2,20 +2,26 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    splited_integer = split_integer(8, 1)
+    assert sum(splited_integer) == 8
```
Iventyk: Due to the checklist, it's a bad way to write tests like this:
You can find the "checklist" file near the README file

Good example:

def test_():
    assert some_function_to_test() == [10]
    
Bad example:

def test_():
    result_of_testing_function = some_function_to_test()
    assert result_of_testing_function == [10]

Iventyk (changes requested):
Hi! Nice job!

But there are some small recommendations :)

app/split_integer.py
Guillelmina8: I was also sure that I didn't need to change the program code itself, but Luke AI said that everything should pass, both the pytest code and my tests. Anyways, thanks for your remark!

app/test_split_integer.py
Guillelmina8: Thank you very much for such a high-quality and detailed comment, I will correct these points immediately ;-)

PR #1765, opened 2025-11-11
Solution

Ann-Riabykina:


tigersaymon (approved):
You don't need to create separate variables for such short cases

PR #1856, opened 2026-01-23
Add tests for split_integer

nazarkravets2007:
Implemented pytest tests for the split_integer function.

Tests cover:
- sum of parts equals the original value
- splitting into equal parts when divisible
- splitting into a single part
- sorted order of parts
- correct behavior when value is less than number of parts


nazarkravets2007 (commented):
Improve test coverage for split_integer

nazarkravets2007 (commented):
Refactor tests to remove unnecessary variables

PR #1972, opened 2026-04-02
solution

Bohdan3457:


app/test_split_integer.py
```diff
@@ -2,20 +2,20 @@
 
 
 def test_sum_of_the_parts_should_be_equal_to_value() -> None:
-    pass
+    assert split_integer(32, 6) == [5, 5, 5, 5, 6, 6]
```
vsmutok:
```suggestion
    assert sum(split_integer(32, 6)) == 32
```
The test name should match what it does.

vsmutok (changes requested):

PR #2006, opened 2026-04-16
Solution

vova-dev-python:


alisahoys (approved):
sum() check is redundant if you already assert exact output.

PR #2114, opened 2026-06-18
Solution

nazarvovchuk5-glitch:


vsmutok (changes requested):
Please do not change the test names.

You should write tests for the exact names that were proposed. Accordingly, each test should verify the behavior or scenario described by its name.

app/test_split_integer.py
```diff
@@ -1,21 +1,65 @@
...
+        split_integer(value, number_of_parts)
+        == sorted(split_integer(value, number_of_parts))
+    )
 
 
-def test_should_add_zeros_when_value_is_less_than_number_of_parts() -> None:
-    pass
+@pytest.mark.parametrize("value, number_of_parts, expected", [
+    (4, 6, [0, 0, 1, 1, 1, 1]), (2, 3, [0, 1, 1]),
+])
```
vsmutok, line 61: Using `@pytest.mark.parametrize` makes sense only when you're testing multiple scenarios. If you intend to test just a single case, remove `@pytest.mark.parametrize`

vsmutok (changes requested):

vsmutok (approved):
Well Done!

PR #2203, opened 2026-08-24
Solution

ArGus20:


Avalanche-one (changes requested):
You don't have to create unnecessary variables for simple tests due to checklist requirements.

Avalanche-one (approved):

